PR #8, opened 2020-03-18
Rename LICENCE to LICENSE

ScottBrenner:
Proposing renaming `LICENCE` to `LICENSE`, quick typo fix.


justincormack:
Hi, this is not a typo, it is UK English. As GitHub recognises both spellings for licence detection I don't think we need to change this.

zappy-shu:
We are meant to be sticking with US English so thanks for your contribution @ScottBrenner

Issue #11, opened 2020-03-18
docs: Change use of "Name" to "Path" in dockerfile option information


matthewfeickert:
Hi. Thanks very much for making this GitHub Action!

I'd like to suggest that in the `dockerfile` option the text

> Name of the Dockerfile

be changed to

> Path to the Dockerfile relative to `path`.

When I first read the docs I got confused by the use of "path" and "name" (instead of "context" and "path") and for a directory that structure that is structured with
```
$ tree
.
├── other_files_needed_for_build
├── docker
│   └── Dockerfile
```
attempted to use

```yaml
    - name: Build and Publish to Registry
      uses: docker/build-push-action@v1
      with:
        username: ${{ secrets.DOCKER_USERNAME }}
        password: ${{ secrets.DOCKER_PASSWORD }}
        repository: org/repo-name
        path: docker
        dockerfile: Dockerfile
```

instead of the following (correct)

```yaml
    - name: Build and Publish to Registry
      uses: docker/build-push-action@v1
      with:
        username: ${{ secrets.DOCKER_USERNAME }}
        password: ${{ secrets.DOCKER_PASSWORD }}
        repository: org/repo-name
        dockerfile: docker/Dockerfile
```

I think the use of "path" would be more clear than "name". If you agree with me I have a branch with this change on my fork that I can open a PR here on (though telling me to do so might involve more typing).

zappy-shu:
Hi @matthewfeickert. Thanks for your feedback.

The README is wrong, the dockerfile path is not changed by the `path` input. I will make a correction.

The `path` input is relates to the `PATH` argument in the `docker build` CLI command while `dockerfile` relates to the `--file` option.

Setting the `path` changes the context as described here: https://docs.docker.com/engine/reference/commandline/build/

matthewfeickert:
Thanks very much, @zappy-shu. I took a look at PR #16 and that seems to perfectly clarify things.

PR #16, opened 2020-03-20
Clarify dockerfile and path inputs

zappy-shu:
Clarify the documentation for the dockerfile and path action inputs including how they relate to each other.

Fixes https://github.com/docker/build-push-action/issues/11

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,13 @@ Example:
 
 ### `path`
 
-Path to run the docker build from. Defaults to `.`.
+Path to the build context. Defaults to `.`
 
 ### `dockerfile`
 
-Name of the Dockerfile. Defaults to `{path}/Dockerfile`.
+Path to the Dockerfile. Defaults to `{path}/Dockerfile`
+
+Note when set this path is **not** relative to the `path` input but is instead relative to the current working directory.
 
 ### `target`
 
--- a/action.yml
+++ b/action.yml
@@ -34,11 +34,11 @@ inputs:
     required: false
     default: false
   path:
-    description: Path to run docker build from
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -86,11 +86,13 @@ Example:
...
 
-Path to run the docker build from. Defaults to `.`.
+Path to the build context. Defaults to `.`
 
 ### `dockerfile`
 
-Name of the Dockerfile. Defaults to `{path}/Dockerfile`.
+Path to the Dockerfile. Defaults to `{path}/Dockerfile`
+
+Note when set this path is **not** relative to the `path` input but is instead relative to the current working directory.
```
justincormack: This is really confusing. If the `path` is `.` then it is relative to the `path`. What do you mean exactly?

justincormack: Do you mean "if you change the default it is relative to the current working directory not the `path`"?

zappy-shu: Yeah pretty much. 

```
path: path/to/wherever
dockerfile: my/Dockerfile
```

Is equivalent to `docker --file my/Dockerfile path/to/wherever` which would mean your directory structure looked something like:

```
.
- my/
  - Dockerfile
- path/
  - to/
     - wherever/
```

"Path to the Dockerfile. Defaults to `{path}/Dockerfile`" Is pretty much a copy/paste of the `docker help build` output:

```
  -f, --file string             Name of the Dockerfile (Default is
                                'PATH/Dockerfile')
```

Happy to change to something more clear if you have better wording

justincormack: I think its ok for now, I think however you explain it it might be confusing...

PR #31, opened 2020-04-16
Adds GCR example

andrew-waters:
Adds an example of using the action with GCR

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,4 +203,5 @@ with:
   password: ${{ secrets.DOCKER_PASSWORD }}
   registry: gcr.io
   repository: myorg/myrepository
+  tags: latest
 ```
--- a/README.md
+++ b/README.md
@@ -193,3 +193,14 @@ with:
   tag_with_ref: true
   target: mytarget
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -192,4 +192,15 @@ with:
   repository: myorg/myrepository
   tag_with_ref: true
   target: mytarget
-```
\ No newline at end of file
+```
+
+The following will build the root Dockerfile, tag the image as `myorg/myrepository:latest`, log in to Google Container Registry using GitHub secrets (where `DOCKER_PASSWORD` is a [JSON key](https://cloud.google.com/container-registry/docs/advanced-authentication#json-key)), and push the image to the CGR repository `myorg/myrepository`:
```
zappy-shu, line 197: I believe unless you specific `tags: latest` or `tag_with_ref: true` then the image will not actually be tagged as anything. This is something that should probably be fixed as defaulting to `{repository}:latest` would be the logical functionality. As it is though could you please correct the example

andrew-waters: Good spot. I've updated it now

Issue #71, opened 2020-07-07
v2: upgrade to buildx


tonistiigi:
relates to https://github.com/docker/github-actions/issues/12

After this action was published, we have been wondering where to take it next. The consensus has been that the best build experience we should recommend should be based on buildx, so we are proposing to refactor the new version of this action to be only based on buildx.

If you are not familiar with buildx https://github.com/docker/buildx, it offers similar UI as old `docker build` command but exclusively works with BuildKit https://github.com/moby/buildkit and offers access to all BuildKit's features, including multi-arch, build-secrets, remote cache etc. and different builder deployment/namespacing options.

Proposal is to have 3 Github actions:

*Build action* - successor (v2) to current build action for invoking a single build. The main difference being that build is executed with Buildx instead of Docker CLI.

*Buildx setup action* - Optional action to install a custom version of buildx/buildkit and/or create a builder instance with the container driver. By default, github environment already provides buildx on the host and docker driver is used. This is somewhat similar to existing community action https://github.com/crazy-max/ghaction-docker-buildx . I've been in contact with @crazy-max to join the efforts under an official action.

*Binfmt action* - Optional action to install binfmt_misc support to host kernel. This is not needed for builds as the latest BuildKit can now do emulation automatically without any setup https://github.com/moby/buildkit/pull/1516 but is needed in case qemu is needed for `docker run` or the user prefers specific qemu version. I have reworked the qemu images we use for different cases in https://github.com/tonistiigi/binfmt that are now fully multi-arch and also produce the artifacts BuildKit can pull for its support. That repo will probably be renamed to something else, and the action's source can live in the same repository.

In addition to new emulation support, some other changes coming in BuildKit include Github token authentication support https://github.com/moby/buildkit/pull/1533 so we can build from git directly, even in private repositories. This improves the cache tracking BuildKit can do on repeated builds and should be the default mode for the action. There is also a PR to enable build secrets from `ENV` that should work better with Github actions where secrets are in `ENV` instead of local files.

[...]

PR #87, opened 2020-08-11
Mono action

crazy-max:
Ref. #71 

## `setup-buildx`

This actions allows to set up Docker [Buildx](https://github.com/docker/buildx). It's written as a [typescript-action](https://github.com/actions/typescript-action/) to be as closed as possible of the [GitHub Runner](https://github.com/actions/virtual-environments) during its execution.

- [x] Allow to choose buildx version (default to the one installed on the runner)
- [x] Allow to install buildx by default
- [x] Set builder name to allow isolated builder instances
- [x] Add [driver input](https://github.com/docker/buildx#--driver-driver). Default to `docker-container`
- [x] Add [driver-opt](https://github.com/docker/buildx#--driver-opt-options) input.
- [x] Add [use](https://github.com/docker/buildx#--use) input. Default `true`
- [x] Output available platforms
- [x] Documentation with examples
- [x] Tests and/or CI workflows

## `setup-qemu`

This action allows to install QEMU static binaries. It's written as a [typescript-action](https://github.com/actions/typescript-action/) to be as closed as possible of the [GitHub Runner](https://github.com/actions/virtual-environments) during its execution.

- [x] QEMU static binaries Docker image. Default to `tonistiigi/binfmt:latest` for now.
- [x] Platforms to install. Default to `all`.
- [x] Output available platforms
- [x] Use it with `setup-buildx` action
- [x] Documentation with examples
- [x] Tests and/or CI workflows

## `build-push`

_Will be made in a further PR_


crazy-max:
@tonistiigi Wonder if **"Opt-in to buildkit for all docker build commands"** is really necessary because buildx will always build using the BuildKit engine and does not require `DOCKER_BUILDKIT=1` environment variable for starting builds.

setup-buildx/.editorconfig
```diff
@@ -0,0 +1,15 @@
+# This file is for unifying the coding style for different editors and IDEs.
```
tonistiigi, line 1: These metadata files probably make more sense in root dir?

setup-buildx/README.md
```diff
@@ -0,0 +1,114 @@
...
+jobs:
+  buildx:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/actions/setup-qemu@v2
```
tonistiigi, line 71: As we are going with the new names, don't think we need v2 for setup-buildx, setup-qemu

setup-buildx/README.md
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          buildx-version: latest
+      -
+        name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
+```
+
+## Customizing
+
+### inputs
```
tonistiigi, line 87: from #71 currently missing load, push, buildkit, driver-opt, use, boot. Fine to leave out from first PR but leave some tracking.

setup-buildx/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `buildx-version` | String  | `latest`  | [Buildx](https://github.com/docker/buildx) version. Example: `v0.3.0` |
+| `install`        | Bool    | `false`   | Sets up `docker build` command as an alias to `docker buildx` |
+
+### outputs
+
+Following outputs are available
+
+| Name          | Type    | Description                           |
+|---------------|---------|---------------------------------------|
+| `platforms`   | String  | Available platforms (comma separated) |
```
tonistiigi, line 102: Builder name output?

setup-buildx/src/main.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    core.info('📣 Buildx info');
+    await exec.exec('docker', ['buildx', 'version']);
+
+    core.info('🔨 Creating a new builder instance...');
+    await exec.exec('docker', [
+      'buildx',
+      'create',
+      '--name',
+      `builder-${process.env.GITHUB_SHA}`,
```
tonistiigi, line 29: I think we should get the name from the step or generate and leave in output. So that `setup-buildx` can be used multiple times if needed.

setup-qemu/src/main.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    const qemuVer: string = core.getInput('qemu-version') || 'latest';
+
+    core.info(`💎 Installing QEMU static binaries...`);
+    await exec.exec('docker', [
+      'run',
+      '--rm',
+      '--privileged',
+      `multiarch/qemu-user-static:${qemuVer}`,
```
tonistiigi, line 19: We should switch to https://github.com/tonistiigi/binfmt releases (the name will probably change at some point). These ones are multi-arch and can be extended with moby patches that we have shipped with Docker desktop before. It is also what buildkit uses by default.

setup-qemu/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Customizing
+
+### inputs
+
+Following inputs can be used as `step.with` keys
+
+| Name             | Type    | Default   | Description                        |
+|------------------|---------|-----------|------------------------------------|
+| `qemu-version`   | String  | `latest`  | [qemu-user-static](https://github.com/multiarch/qemu-user-static) version (Docker tag). Example: `4.2.0-7` |
```
tonistiigi, line 46: https://github.com/tonistiigi/binfmt supports setting the needed architectures and querying what emulators are enabled for the outputs.

setup-buildx/src/installer.ts
```diff
@@ -0,0 +1,66 @@
...
+import * as tc from '@actions/tool-cache';
+
+const osPlat: string = os.platform();
+
+export async function buildx(inputVersion: string, dockerConfigHome: string): Promise<string> {
+  const release: github.GitHubRelease | null = await github.getRelease(inputVersion);
+  if (!release) {
+    throw new Error(`Cannot find buildx ${inputVersion} release`);
+  }
+  core.debug(`Release found: ${release.tag_name}`);
```
tonistiigi, line 17: Do I understand this correctly that if buildx release already exists in the GH environment then the download is skipped?

setup-buildx/README.md
crazy-max: It's already annotated as v2 `...@v2` or I missed something there?

setup-buildx/README.md
crazy-max: Ok will add them in the todo of the PR for now and start working on it.

.github/workflows/setup-qemu-ci.yml
```diff
@@ -16,9 +16,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        qemu-version:
-          - latest
-          - 4.2.0-7
+        image:
+          - tonistiigi/binfmt:latest
+          - tonistiigi/binfmt:buildkit
```
tonistiigi, line 21: `:buildkit` version is not to be used in kernel. we can test `:master` if you like. Can also build for specific qemu versions but I think atm there is no need to ever use old version.

setup-buildx/src/installer.ts
crazy-max: In this case no, if the action doesn't find a buildx version available it throws an exception. So it might be wise to raise only a warning informing the user that the default version of buildx available on the runner will be used?

setup-buildx/src/main.ts
crazy-max: Here we use the `GITHUB_SHA` env var as it's unique for the **current job** of the workflow. This has been implemented through crazy-max/ghaction-docker-buildx#116 because some users have their own runner and multiple workflows can run on one runner.

If we want multi builders per job we can append the name to the job SHA to avoid conflicts. WDYT?

setup-buildx/README.md
crazy-max: Will add it. See also https://github.com/docker/build-push-action/pull/87#discussion_r468847851

setup-buildx/README.md
crazy-max: * I think `load` and `push` inputs has to be implemented through the `build-push` action not `setup-buildx` as it's only available through `docker buildx build` command right?
* About the `buildkit` input I don't think it's really necessary because buildx will always build using the BuildKit engine and does not require `DOCKER_BUILDKIT=1` environment variable for starting builds. Or maybe you talk about the `moby/buildkit` Docker image tag version for this one?

setup-buildx/src/installer.ts
tonistiigi: I think the desired ux would be that download is skipped by default because github env already ships with buildx. If user sets a manual version we override the default buildx with a downloaded binary. Could also check if the default version was already the same that user requested but that level of optimization isn't really important.

setup-buildx/src/main.ts
tonistiigi: > because some users have their own self-hosted runner and multiple workflows can run on one runner.

That seems very error-prone. If they let the jobs collide the built (local) image names will also collide and provide wrong results. If they know they have some weird limits like this can't they just assign that random component themselves with an input.

Also, how is `GITHUB_SHA` unique? You could also have many jobs running for the same sha that will still cause a collision. Then It would need to be actual random component.

At least lets make sure to provide the name in the output.

setup-buildx/README.md
tonistiigi: > I think load and push inputs has to be implemented through the build-push action not setup-buildx as it's only available through docker buildx build command right?

I was thinking this could provide a default behavior so it doesn't need to be set for every build. Lmk if you would prefer some buildx changes for this (or if you just think it is a bad idea). We have 2 groups of users, ones that want the new isolated multi-arch experience and others who just want updated local `docker build` experience. Trying to find a way to support both.

> Or maybe you talk about the moby/buildkit Docker image tag version to use with driver-opt for this one?

Yes, this is what I had in mind. Eg. if you want to use master version with latest updates. If we expose full driver-opt maybe there is no need for a special input for buildkit image.

setup-buildx/.editorconfig
tonistiigi: What about the other dotfiles?

setup-buildx/README.md
crazy-max: > I was thinking this could provide a default behavior so it doesn't need to be set for every build. Lmk if you would prefer some buildx changes for this (or if you just think it is a bad idea). We have 2 groups of users, ones that want the new isolated multi-arch experience and others who just want updated local docker build experience. Trying to find a way to support both.

Both seem feasible and not necessarily contradictory to me. So ok to add these 2 inputs and also allow to override that through the future `build-push` action.

> Yes, this is what I had in mind. Eg. if you want to use master version with latest updates. If we expose full driver-opt maybe there is no need for a special input for buildkit image.

Ok I have added `driver` and `driver-opt` inputs but for me it looks like `buildkit` input is redundant and could mislead the user experience compared to the already existing input `buildx-version`.

setup-buildx/src/installer.ts
crazy-max: Ok done. Now I check if buildx is not available on runner or `buildx-version` setted to install it.

setup-buildx/src/main.ts
crazy-max: > That seems very error-prone. If they let the jobs collide the built (local) image names will also collide and provide wrong results. If they know they have some weird limits like this can't they just assign that random component themselves with an input.

I want to at least make sure that the action does not raise an exception when creating the builder instance. About local image build collision and furthermore self-hosted runner we can discuss about it in the next `build-push` action PR?

> Also, how is `GITHUB_SHA` unique? You could also have many jobs running for the same sha that will still cause a collision. Then It would need to be actual random component.

After digging, `GITHUB_SHA` is not unique per job but by event, so yes, a collision with builders can happen. I will use the `INVOCATION_ID` env var which is unique per job.

> At least lets make sure to provide the name in the output.

:+1:

crazy-max: Done: https://github.com/docker/build-push-action/runs/977235076?check_suite_focus=true

setup-buildx/README.md
crazy-max: Done: https://github.com/docker/build-push-action/runs/977235076?check_suite_focus=true

setup-buildx/src/main.ts
```diff
@@ -0,0 +1,73 @@
...
+import * as path from 'path';
+import * as buildx from './buildx';
+import * as exec from './exec';
+import * as stateHelper from './state-helper';
+import * as core from '@actions/core';
+
+async function run(): Promise<void> {
+  try {
+    if (os.platform() !== 'linux') {
+      core.setFailed('Only supported on linux platform');
```
tonistiigi, line 11: Not sure if this is a good idea but we do have mac and windows releases of buildx. It would not work by default though but with a remote endpoint. Thinking about that, maybe we should support remote endpoint, but don't have a good flow in mind atm. Maybe option that allows skipping "create" step so create can be run inside workflow with custom endpoint configuration.

setup-qemu/.gitattributes
```diff
@@ -0,0 +1,2 @@
+/dist/** linguist-generated=true
```
tonistiigi, line 1: these 3 dotfiles are still per subfolder and not per-repo. Maybe you are waiting on the solution for the naming question first. Ignore then.

setup-qemu/README.md
```diff
@@ -0,0 +1,64 @@
...
+## Customizing
+
+### inputs
+
+Following inputs can be used as `step.with` keys
+
+| Name             | Type    | Default                     | Description                        |
+|------------------|---------|-----------------------------|------------------------------------|
+| `image`          | String  | `tonistiigi/binfmt:latest`  | QEMU static binaries Docker image. e.g. [`tonistiigi/binfmt:latest`](https://hub.docker.com/r/tonistiigi/binfmt/tags) |
+| `platforms`      | String  | `all`                       | Platforms to install. e.g. `arm64,riscv64,arm` |
```
tonistiigi, line 52: Could we list all supported values.

setup-qemu/.gitattributes
crazy-max: > Maybe you are waiting on the solution for the naming question first. Ignore then.

Yes, I prefer to wait for our discussion on the subject before making these changes.

setup-buildx/src/main.ts
crazy-max: I don't think that adding support for platforms other than Linux is paramount as it stands, especially since Docker is not installed on [MacOS virtual environments](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md).

setup-qemu/README.md
crazy-max: You mean all available platforms? If that's the case we should change the README "often" enough. Maybe we should rather point to a link listing the available platforms?

tonistiigi: Yeah, link is good.

setup-buildx/src/main.ts
tonistiigi: Yeah, not that important. Buildx does work without a docker binary as well though.

crazy-max: I thought of adding a link to the [containerd database platforms](https://github.com/containerd/containerd/blob/master/platforms/database.go) but it seems confusing to me. Do you have a link with a list of potentially supported platforms?

setup-qemu/README.md
crazy-max: I thought of adding a link to the [containerd database platforms](https://github.com/containerd/containerd/blob/master/platforms/database.go) but it seems confusing to me. Do you have a link with a list of potentially supported platforms?

PR #92, opened 2020-08-15
Build Push action v2

crazy-max:
Ref. #71 

v2 of Build Push action allows to use Docker [Buildx](https://github.com/docker/buildx). It's also rewritten as a [typescript-action](https://github.com/actions/typescript-action/) to be as closed as possible of the [GitHub Runner](https://github.com/actions/virtual-environments) during its execution.

- [x] Rename `path` input to `context`
- [x] Rename `dockerfile` input to `file`
- [x] Rename `always_pull` input to `pull`
- [x] Add `builder` input to be able to choose a builder instance through **setup-buildx action**
- [x] Add [`platforms`](https://github.com/docker/buildx#---platformvaluevalue) input
- [x] Add [`allow`](https://github.com/docker/buildx#--allowentitlement) input
- [x] Add [`load`](https://github.com/docker/buildx#--load) input
- [x] Add [`outputs`](https://github.com/docker/buildx#-o---outputpath-typetypekeyvalue) input
- [x] Add [`cache-from`](https://github.com/docker/buildx#--cache-fromnametypetypekeyvalue) input
- [x] Add [`cache-to`](https://github.com/docker/buildx#--cache-tonametypetypekeyvalue) input
- [x] Add `secrets` input
- [x] Review `tags` input
- [x] Remove `cache_froms` input
- [x] Handle Git context
- [x] Add `digest` output
- [x] Move login support in a dedicated action (see below)
- [x] Documentation with examples
- [x] Fix push on local registry on CI workflow (needs #93 to be merged)
- [x] Tests and/or CI workflows

## Other features (maybe future?)

- Review `tag_with_sha`, `tag_with_ref`, `add_git_labels` inputs
- Handle [OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/master/annotations.md)
- **Done** ; see https://github.com/docker/build-push-action/pull/92/commits/44d05b9191c2810e61484c43e6fe4b653e97cf2a ; Leverage GitHub cache across jobs for buildx through the [@actions/cache](https://github.com/actions/toolkit/tree/master/packages/cache) package
- **Done** ; see https://github.com/docker/build-push-action/pull/92/commits/2962fe9789cfac6a55f4ab647858dfe76d265069 ; Handle [bake](https://github.com/docker/buildx#buildx-bake-options-target)

## Docker login

As stipulated during our Zoom session, the authentication part has been removed and will be handled by a dedicated action. This will allow actions such as [Azure Container Registry Login](https://github.com/marketplace/actions/azure-container-registry-login) or [Amazon ECR "Login"](https://github.com/marketplace/actions/amazon-ecr-login-action-for-github-actions) to work efficiently with it. I have also created an action allowing authentication with several examples to authenticate against registries. We will be able to move the repository within the organization if it suits you (repo will be named **docker/login-action** for example):

* https://github.com/marketplace/actions/docker-login
* https://github.com/crazy-max/ghaction-docker-login

## Deprecation notice

It would be advisable to set up a deprecation notice on the action currently in place in order to inform users of a new major version. This can be done directly in the `action.yml` file [like this](https://github.com/Azure/k8s-actions/blob/7a93ceb73dcebce42afdbb13e5eca8548143a281/docker-login/action.yml#L8).


crazy-max:
@tonistiigi I've done some tests for using git via buildkit and it turns out that implementing it in our action would add a layer of complexity that I think should be delegated to [actions/checkout](https://github.com/actions/checkout/). As you can see in their [Usage section](https://github.com/actions/checkout/#usage), they already managed rough cases like credentials persistance, submodules, token, ssh key. Their [code base](https://github.com/actions/checkout/tree/main/src) is also quite huge to handle all the special cases so I wonder if we should really implement and maintain this layer in our action.

tonistiigi:
@crazy-max We shouldn't do any of these complicated cases in our action code. But buildkit does have first-class support for building from git (as did old docker build) and we should just reuse it. There is no need to have any kind of compatibility with all the `actions/checkout` options. People can still use `actions/checkout` and set the context path to `.`. From everything you listed `ssh-key` is not currently  supported but will probably be in the future https://github.com/moby/buildkit/issues/1564 . The main issue was with the host key validation, probably the same kind of extra fields would be needed that `actions/checkout` has added.

There are many benefits for building from git. Eg. even in current BuildKit CI that still uses travis, even though travis automatically does a checkout, we ignore it and just pass the git ref from `ENV` directly to buildkit instead.

crazy-max:
@tonistiigi 

> People can still use `actions/checkout` and set the context path to `.`

Ok LGTM. I will add an example and tests using git and the github token with the current implementation.

tonistiigi:
@crazy-max So context path can be set to local path, git path or HTTP(s) path. If git path is set, expose token as well. If no value is set then it defaults to the git path that is determined from the environment variables.

crazy-max:
@tonistiigi Git context [works fine](https://github.com/docker/build-push-action/runs/990902891?check_suite_focus=true#step:6:21). I have [added an example](https://github.com/docker/build-push-action/tree/v2-build-push#git-context) in the README using the [GitHub context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) and a note about the [available environment variables](https://github.com/docker/build-push-action/tree/v2-build-push#environment-variables).

tonistiigi:
@crazy-max In the example, I'm not sure if you just want to show how to override values or if these inputs are not optional. Fields like `GIT_AUTH_TOKEN`, `GIT_REF`, `context: ${{ github.repositoryUrl }}` should not be needed for default behavior(not `builder:` but that is separate). If they are already optional then I think we should have more examples for different cases(simple default flows, more complicated ones with overrides etc), but these changes can be in follow-up.

Another question I had was if type of input needs to be fixed to either string or a list? If I remember correctly it does not. I think this would help in cases for example for `platforms` field that could be either a string or a list of strings.

crazy-max:
@tonistiigi

> In the example, I'm not sure if you just want to show how to override values or if these inputs are not optional. Fields like `GIT_AUTH_TOKEN`, `GIT_REF`, `context: ${{ github.repositoryUrl }}` should not be needed for default behavior(not `builder:` but that is separate).

Yes, that's an example to leverage all available options if using git context.

> If they are already optional then I think we should have more examples for different cases(simple default flows, more complicated ones with overrides etc), but these changes can be in follow-up.

Context is required if you want git context enable because for now default context is `.`. So from what I see you want git as our default context?

> Another question I had was if type of input needs to be fixed to either string or a list? If I remember correctly it does not. I think this would help in cases for example for `platforms` field that could be either a string or a list of strings.

Yes, I wondered about that as well. I generally prefer inputs delimited by a new line in order to have a better reading of the workflow. But I can handle both cases very well with a slight modification of [this function](https://github.com/docker/build-push-action/blob/6ede7e932d90ffa0c9375eb3b6f6cbc4c4d1bb8f/src/context.ts#L148-L154).

crazy-max:
@tonistiigi I think it's a bit out of our scope but I've implemented a specially designed cache through GitHub Actions through their official  [@actions/cache](https://github.com/actions/toolkit/tree/master/packages/cache) package. For this I have only set up a new `cache-github` boolean input which will set the values of `--cache-from` and `--cache-to flags`.

I think that the added value of its implementation will allow us to benefit from a cache through several jobs executed on the repository without having to push the cache on a registry. Especially since this cache is quite fast through the Azure Blob Storage (see https://github.com/docker/build-push-action/runs/991261326?check_suite_focus=true#step:5:18).

Let me know if you're interested.

tonistiigi:
> Yes, I wondered about that as well. I generally prefer inputs delimited by a new line in order to have a better reading of the workflow. But I can handle both cases very well with a slight modification of this function.

Yeah, I think it would be good to do that and change `platforms` and `allow` to a list.

> but I've implemented a specially designed cache through GitHub Actions with their official @actions/cache package.

I'll take a look. My main question is if this should be implemented in buildkit as a cache backend instead. The benefit of that is that buildkit can determine a cache match before pulling anything and only pull matched data. This allows the cache to grow larger without builds getting slower. Also I think this would handle cache cleanups better as blobs would be cached separately so they also get cleaned up separately and you don't lose the whole cache.

I also saw you added bake keys. Not sure atm if that is a good idea and probably need more opinions on this. As single build and bake are quite different beasts and configured with different inputs I'm not sure if it makes sense to handle them in a same action. Maybe for this one it makes sense to use subaction of the build action?

crazy-max:
> Yeah, I think it would be good to do that and change `platforms` and `allow` to a list.

[Done](https://github.com/docker/build-push-action/pull/92/commits/363c8ed24f1fd27f60a8d58e2c54cb34e7ad2820).

> I'll take a look. My main question is if this should be implemented in buildkit as a cache backend instead(I assume atm it uses the local backend).

Yes indeed the GitHub cache uses the local backend.

> The benefit of that is that buildkit can determine a cache match before pulling anything and only pull matched data. This allows the cache to grow larger without builds getting slower. Also I think this would handle cache cleanups better as blobs would be cached separately so they also get cleaned up separately and you don't lose the whole cache.

Yes that sounds like a good idea! In the long term I think we should add an Azure Blob Storage backend and use the workflow [context information (`ACTIONS_CACHE_URL`)](https://github.com/actions/toolkit/blob/master/packages/cache/src/internal/cacheHttpClient.ts#L33-L47.) to determine if this backend can be used. And ideally use the [Azure Storage Blob SDK for Go](https://github.com/Azure/azure-storage-blob-go) for this backend like the [official GitHub Action cache lib](https://github.com/actions/toolkit/blob/1cc56db0ff126f4d65aeb83798852e02a2c180c3/packages/cache/src/internal/downloadUtils.ts#L205-L224).

crazy-max:
@tonistiigi Looks like [subdir is not yet supported](https://github.com/docker/build-push-action/runs/994283076?check_suite_focus=true#step:6:28) through git context? `git://github.com/docker/build-push-action.git#refs/pull/92/merge:test`

tonistiigi:
@crazy-max Indeed. Is there a way the subdir case can be reached in the default path where git context is determined from the env vars?

crazy-max:
@tonistiigi 

> Ok LGTM. I will take a look if I could leverage the new composite feature to handle this.

Composite does not accept `uses` step so we should create a subaction. In the meantime I will remove the `bake` support for this action and we could do this in a future release.

> Indeed. Is there a way the subdir case can be reached in the default path where git context is determined from the env vars?

If we can rely on an environmental variable, yes, I think it would be possible.

crazy-max:
@tonistiigi Ready to be reviewed. I have also updated my [test-docker-action repo](https://github.com/crazy-max/test-docker-action) and included our [3 actions](https://github.com/crazy-max/test-docker-action/runs/998377465?check_suite_focus=true).

@justincormack Maybe it's time to create the two repos **setup-buildx-action** and **setup-qemu-action** in order to migrate actions before merging?

Let me know if you need anything.

alexellis:
@crazy-max given that this PR is still pending, what action would you suggest using that supports buildx in the meantime?

crazy-max:
Hi @alexellis, in the meantime you can use the [following workflow](https://github.com/crazy-max/test-docker-action/blob/master/.github/workflows/ci.yml). If you encounter any issue, do not hesitate to bring them up here.

src/context.ts
```diff
@@ -0,0 +1,127 @@
...
+  push: boolean;
+  outputs: string[];
+  cacheFrom: string[];
+  cacheTo: string[];
+}
+
+export async function getInputs(): Promise<Inputs> {
+  return {
+    context: core.getInput('context') || '.',
+    file: core.getInput('file') || './Dockerfile',
```
RammusXu, line 32: I suppose it will be better UX to use `$context/Dockerfile` as default.

tonistiigi: the context default should be git url and default file `Dockerfile` inside it

RammusXu: It will be nice If compatible with the [build-push-action@v1 logic](https://github.com/docker/build-push-action#dockerfile)

also similar to [docker build](https://docs.docker.com/engine/reference/commandline/build/#options)
```
docker build [OPTIONS] PATH | URL | -
--file , -f |   | Name of the Dockerfile (Default is ‘PATH/Dockerfile’)
```

tonistiigi: It will still be somewhat compatible as if you use `actions/checkout` they point to the same source. This change is one of the reasons for v2.

crazy-max: @tonistiigi After digging to default the context to git, some events have an unqualifed ref. For example when a PR is merged (pull_request closed event), the ref is unqualifed like "master" instead of "refs/heads/master" so it could break a build. GitHub has [some code](https://github.com/actions/checkout/blob/main/src/ref-helper.ts) to deal with this in their checkout action but I don't think we should impl/maintain this behavior here as it would not be cost effective in the future. WDYT?

tonistiigi: build should already support both cases as well (it should support all refspec git supports) . If there is a case that is not currently supported it should be added as building from git is an important feature and should work from any ref.

crazy-max: @tonistiigi Ok Git context is the default one now. I have also added support for secrets through a `secrets` input.

src/buildx.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return fs.readFileSync(iidFile, {encoding: 'utf-8'});
+}
+
+export async function getSecret(kvp: string): Promise<string> {
+  const [key, value] = kvp.split('=');
+  const secretFile = tmp.tmpNameSync({
+    tmpdir: context.tmpDir
+  });
+  await fs.writeFileSync(secretFile, value);
```
tonistiigi, line 25: what is removing this file when action completes?

src/main.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    stateHelper.setTmpDir(context.tmpDir);
+
+    const buildxVersion = await buildx.getVersion();
+    core.info(`📣 Buildx version: ${buildxVersion}`);
+
+    let inputs: context.Inputs = await context.getInputs();
+    if (inputs.builder) {
+      core.info(`📌 Using builder instance ${inputs.builder}`);
+      await buildx.use(inputs.builder);
```
tonistiigi, line 28: this should probably use `--builder` flag in buildx. Otherwise, the default builder will switch for next commands and that might be unexpected.

tonistiigi: buildx also understands `BUILDX_BUILDER` env in case that is easier to plug in.

src/buildx.ts
crazy-max: It is done in the [post step](https://github.com/docker/build-push-action/blob/b563df09ad646120c1d28138a0f9fa687817885e/src/main.ts#L46-L51)

src/main.ts
crazy-max: Done

sualeh:
@crazy-max - thanks for making the changes. The way I read this, you have a separate buildx action, and then reference it in the build-push action. What Id I don't use buildx, and build using some other means - how can I say that the build is done locally, just push it?

tonistiigi (approved):

crazy-max:
Mutable tag created: https://github.com/docker/build-push-action/tree/v2
Please use `docker/build-push-action@v2` for those who try the new action. `v2-build-push` branch will be deleted as the PR has been merged.

PR #93, opened 2020-08-16
Fix and cleanup of v2 setup actions

crazy-max:
Ref. #71

* Cleanup unused deps
* Fix package/action info
* Fix post state builder name for setup-buildx
* Add `buildkitd-flags` input


setup-buildx/README.md
```diff
@@ -91,13 +91,14 @@ jobs:
...
-| `install`        | Bool    | `false`                           | Sets up `docker build` command as an alias to `docker buildx` |
-| `use`            | Bool    | `true`                            | Switch to this builder instance |
+| Name               | Type    | Default                           | Description                        |
+|--------------------|---------|-----------------------------------|------------------------------------|
+| `version`          | String  | _the one installed on the runner_ | [Buildx](https://github.com/docker/buildx) version. e.g. `v0.3.0`, `latest` |
+| `driver`           | String  | `docker-container`                | Sets the [builder driver](https://github.com/docker/buildx#--driver-driver) to be used. |
+| `driver-opt`       | String  |                                   | Passes additional [driver-specific options](https://github.com/docker/buildx#--driver-opt-options). e.g. `image=moby/buildkit:master` |
+| `buildkitd-flags`  | String  |                                   | [Flags for buildkitd](https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md) daemon. |
```
tonistiigi, line 99: wondering if should add special case for `allow-insecure-entitlement` so it works together with `allow` in build action. Or actually, maybe we should just always enable `allow-insecure-entitlement` in github actions context as these run in a isolated VM anyway. I could also add `--allow-insecure-entitlement=all` in buildkit.

crazy-max: @tonistiigi 

> wondering if should add special case for allow-insecure-entitlement so it works together with allow in build action. Or actually, maybe we should just always enable allow-insecure-entitlement in github actions context as these run in a isolated VM anyway.

Yes, that was the purpose of integrating buildkit-flags so that I can use a [local registry for testing](https://github.com/docker/build-push-action/runs/989247326?check_suite_focus=true#step:6:197). And indeed in the case of GitHub Actions it seems logical to me to [enable `allow-insecure-entitlement` by default](https://github.com/docker/build-push-action/runs/989720106?check_suite_focus=true#step:6:20).

> I could also add --allow-insecure-entitlement=all in buildkit.

Yes SGTM!

crazy-max: @tonistiigi Ok local registry [is working](https://github.com/docker/build-push-action/runs/991623873?check_suite_focus=true#step:7:1) in our [ci workflow](https://github.com/docker/build-push-action/blob/8413351ed4d00d8a81c936446c73807def589b75/.github/workflows/ci.yml#L35-L58).

PR #104, opened 2020-09-05
Local registry example

crazy-max:


README.md
```diff
@@ -221,6 +211,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
```
tonistiigi, line 236: Lets skip the things that are not needed for this example

README.md
```diff
@@ -221,6 +211,54 @@ jobs:
...
+          driver-opts: network=host
+          #buildkitd-flags: --allow-insecure-entitlement security.insecure # default
+      -
+        name: Build and push to local registry
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/386,linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
+          allow: network.host,security.insecure
```
tonistiigi, line 253: skip

tonistiigi (commented):
Isn't setup-buildx action better for this example?

crazy-max:
@tonistiigi 

> Isn't setup-buildx action better for this example?

I think in most cases the use of a local registry will involve pushing an image into it so in my opinion this example is better suited here.

README.md
tonistiigi: this has not been addressed. allow shouldn't be required for this. As well as `buildkitd-flags`

crazy-max: Done

tonistiigi (approved):

PR #106, opened 2020-09-08
Add deprecation notice (v2 available)

crazy-max:
I choose the `repository` input because it's the only required one and is not used in our v2.


crazy-max:
Will look [like this](https://github.com/docker/build-push-action/actions/runs/244987815) for the end user:

image

tonistiigi (approved):

PR #117, opened 2020-09-09
Add example to update DockerHub repo description

### LINKED ISSUE #21: Update Readme and description with push on Docker hub
Currently, if we use the Github action with this repo, the image is built and pushed to the Docker hub, but the Readme and description are not updated on the Docker hub webpage. It would be nice to have this in the future to use this action and altogether disable auto build.

crazy-max:
Closes #21 

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ ___
   * [Local registry](#local-registry)
   * [Leverage GitHub cache](#leverage-github-cache)
   * [Complete workflow](#complete-workflow)
+  * [Update DockerHub repo description](#update-dockerhub-repo-description)
 * [Customizing](#customizing)
   * [inputs](#inputs)
   * [outputs](#outputs)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -365,6 +366,50 @@ jobs:
...
+        name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: user/app:latest
+      -
+        name: Update repo description
+        uses: peter-evans/dockerhub-description@v2
+        env:
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
```
felipecrs, line 408: Isn't it the same user and password? I think you should reuse the secrets.

crazy-max: @felipecrs As stipulated by the Docker Hub Description action, you cannot use a DockerHub Token to update the description. That's why I don't use the same credentials here. More info: https://github.com/peter-evans/dockerhub-description#usage

felipecrs: Maybe the `DOCKER_PASSWORD` should be renamed as `DOCKER_TOKEN` so :)

crazy-max: @felipecrs `DOCKER_PASSWORD` can be used as `docker login` but yeah good idea to use `DOCKER_TOKEN` as best practice.

PR #119, opened 2020-09-10
Add tags and labels handling example

### LINKED ISSUE #13: Automatic creation of staggered versioned tags
I'd would be cool if there was a option for this action to automatically split the version of the git reference version up and create multiple docker tags based on the major and minor version numbers of the git reference.

For example: `refs/tags/v1.3.1` would turn into

- `v1.3.1`
- `v1.3`
- `v1`

Sorry don't know if there is a proper name for this practice 😅 But this seems to be common among the more popular images.

### LINKED ISSUE #23: Add tag_prefix input
I think an input for `tag_prefix` would be very useful in some usecases. If given, this value would be prefixed to all tags on on the image. With the current inputs, the `tag_prefix` would be prefixed to the following tag inputs. Note that I've included `tags` here as wel as I think that would be most consistent, but that is the one I'm the least sure about whether that would make sense.

- [`tags`](https://github.com/docker/build-push-action#tags)
- [`tag_with_ref`](https://github.com/docker/build-push-action#tag_with_ref)
- [`tag_with_sha`](https://github.com/docker/build-push-action#tag_with_sha)

One example, which is my usecas for this input and is something the official [ubuntu image](https://hub.docker.com/_/ubuntu) tags reflect as well.

When we have one GitHub repository with multiple Dockerfiles ([see mine as example here](https://github.com/vidavidorra/docker-linux/tree/master/docker)) to build different images (Ubuntu versions in my case). Let's say we have the following images we want to build.
- `bionic`
- `xenial`
- `focal`

WIth the `tag_prefix` with the values of the list above and e.g. a SHA of `abc1234`, these images would get the following tags.
- `bionic-sha-abc1234`
- `xenial-sha-abc1234`
- `focal-sha-abc1234`

I hope I explained everything in here correctly, but I'd be happy to elaborate and discuss.

Edit:
I've locally tested this and it should work as described above.

[input_tag_prefix.patch.txt](https://github.com/docker/build-push-action/files/4398986/input_tag_prefix.patch.txt)

### LINKED ISSUE #33: Feature request: Timestamp based tag
I would like to have some time stamp based tag - built in. Can be done using `(date +%s)` or similar.
I could try to contribute it myself, but I couldn't find out where the actual script that run :(

### LINKED ISSUE #40: Enable use of non-master branch as "latest"
From the `tag_with_ref` docs: 

> If {branch-name} is master then the tag will be latest.

We use a workflow that has separate `master` and `dev` branches, with strict rules around the interaction between them. Anything that requires a rebuild of the image (i.e. updating dependencies, rather than source files) must go through `dev` first. We have set `dev` as our main branch in Github, and it's locked, with controls around how devs can push to it.

We would like to have the `latest` tag always applied to the `dev` branch.

Appreciate we may be outliers here, but if it was possible to add a setting, it would really make a difference for us. I'll dig around myself and if I can see how to do it, will submit a PR.

### LINKED ISSUE #42: Remove `v` from tags
I think the action should have the option to remove the `v` from git tags that are semver. And I think it should be the default behavior.

Git tag: `v1.1.1` -> Docker tag: `1.1.1`

### LINKED ISSUE #62: [Feature] Pass tags and labels as build args as well
The action computes lots of useful values, for example it creates tags using options like `tag_with_ref ` or `tag_with_sha`; and it adds many useful labels with `add_git_labels`

## Feature Request
What's missing, however, is an option that allows passing all those values as build arguments, so we can make use of them in our dockerfile.

Since all the information is already there, the feature boils down to just adding one or two more options:

- Maybe a simple bool  
  `pass_git_labels_as_build_args: true` (or maybe something shorter :P)
- Or maybe allow users to select what to pass  
   `computed_build_args: [ "git_ref", "git_sha", "org.opencontainers.image.created" ]`

## Use case
Most applications need to know their version/build time, so they can display it somewhere (in the header of their commandline output, or "about box", ...)

### LINKED ISSUE #66: Pushing with image variants
I would like, in addition to tags, add a `variant` part to docker image, like:
* `metacontrollerio/metacontroller:v4.2.0-alpine`
* `metacontrollerio/metacontroller:v4.2.0-debian`

etc. What is the easiest way to do it ?

### LINKED ISSUE #83: Append build target to tag_with_ref
If I understand the `tag_with_ref` documentation correctly, it doesn't not take the `target` into account. Therefore, containers with different build targets will override each other. For example, I can't use `tag_with_ref` for a workflow where I first build and deploy a container for the development environment, then run my unit tests in it, and then build and deploy the application.

I suggest you add an option to append the `target` to the `tag_with_ref`, or do that automatically.

### LINKED ISSUE #96: Unsupported chars in PR/Branch name when using tag_with_ref
This is more of a query - is there an easy way to 'clean up' the reference used for `tag_with_ref` before building?
e.g. When pushing from a branch - branch name has a '#' in it (as it is referencing GitHub issue number).

This fails when building the image as the '#' is not a valid tag character.
Obviously we can work around it by changing branch name formats but it would be nice to be able to strip out invalid chars if someone did inadvertently use non-tag friendly chars. Or do we need to manually build up a tag ourselves and use `tag` instead of `tag-with-ref`?

### LINKED ISSUE #101: PR tag is not being set accordingly
In my builds, the PR tag is being set to something like:

`pr-14-merge`

See: https://github.com/felipecrs/jenkins-agent-dind/runs/1072507392?check_suite_focus=true#step:6:21

It should be set to `pr-14` only. This is probably because the build was triggered by a pull-request event and not a regular push, but the action should support that case.

### LINKED ISSUE #116: Missing high-level features from v1 in v2 like `tag_with_ref` and `add_git_labels`
The utility in using the action to me, over just running the `docker` commands manually, was in the high-level interface offered by helpers in `v1` like:

* `add_git_labels`
* `tag_with_sha`
* `tag_with_ref`

These all appear to have disappeared in `v2`. Are these planned to be brought back?

Doing:

```yaml
-  name: Build and push
   with:
     push: true
     tags: user/app:latest
```

Doesn't provide a huge benefit over:

```yaml
- name: Build and push
  run: |
     docker buildx build --push user/app:latest .
```

Compared to implementing any of the following which requires lots of logic and loops.

* `add_git_labels`
* `tag_with_sha`
* `tag_with_ref`

crazy-max:
Closes #13
Closes #23
Closes #33 
Closes #40
Closes #42
Closes #62
Closes #66
Closes #83
Closes #96
Closes #101
Closes #116 


tonistiigi:
This looks quite a crazy hack. Maybe if we want to provide a workaround before https://github.com/docker/build-push-action/issues/116#issuecomment-690531322 we could link to another action that does this and sets the output?

metcalfc:
It doesn't look crazy. It looks like a shell script. I think its a reasonable workaround for folks that need the behavior from v1 to try v2. If there is an issue on deck in buildkit to get this functionality soon why burn cycles on a javascript action? Now if that issue in buildkit isn't coming any time soon because its not scoped or scheduled... Well then yes we should probably add something because I think we need that functionality for v2 GA. 

The downside is once we add the an action folks are going to use it and its going to be around for a lot longer then a bash snippet.

docwhat (approved):

crazy-max:
@tonistiigi 

> Maybe if we want to provide a workaround before #116 (comment) we could link to another action that does this and sets the output?

Yes we could create a dedicated action to handle that or like I said to you on Slack we could use a "magic word" in `tags` input to generate tags based on git reference [like this](https://github.com/crazy-max/docker-build-push-action/blob/bc0ff20cc0e14d78343772f2b36cbbe39d817f33/src/context.ts#L145-L185). WDYT?

tonistiigi:
I didn't necessarily think we need to create a special action for this ourselves. Was more thinking if we should spend so much time explaining a complicated edge case instead of linking to an implementation (like we do for hub readme update for example). Obviously nobody is stopping anyone to use script like this in their workflow if they like it. Although this is linked to a lot of issues the fact that this has not been brought up in `docker build` for 7 years shows it is not that common. Also, I'd suggest in most cases the complicated script is overkill and users can just use a constant or variable replacement directly in their workflow. Anyway, I'm not against merging this to show examples of workaround. Maybe explain a bit more that this is an example that shows many cases that you might want to do. Eg. breaking up the semantic version into list of tags may be useful for some projects but looks quite unlikely that the same step would want to push every pr by its number as well.

crazy-max:
@tonistiigi 

> Maybe explain a bit more that this is an example that shows many cases that you might want to do.

Ok LGTM I will enhanced the README about this and create an "advanced usage" section.

> Eg. breaking up the semantic version into list of tags may be useful for some projects but looks quite unlikely that the same step would want to push every pr by its number as well(not sure if pushing from pr is even possible due to auth).

Indeed, [secrets are not passed to the runner when a workflow is triggered from a forked repository](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) so  pushing from pr is not possible. I will update the README about that too.

tonistiigi (approved):

PR #123, opened 2020-09-13
Add support to pass `--ssh` flag to the build

### LINKED ISSUE #39: Feature request: provide way to add --ssh argument to docker build command
There does not appear to be a way to turn on the newish --ssh forwarding capability of docker build with this action.   This is needed to access private github repos from inside the Dockerfile (etc).  If this is in some way available, it is not documented.

Some way to place the ssh key(s) (or inject them from a repo secret) and get them to work with RUN --mount=type=ssh would be great.  If this ability is already present, then documentation would be greate!

jesserockz:
Closes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -673,6 +673,7 @@ Following inputs can be used as `step.with` keys
 | `cache-from`        | List     | List of [external cache sources](https://github.com/docker/buildx#--cache-fromnametypetypekeyvalue) (eg. `type=local,src=path/to/dir`) |
 | `cache-to`          | List     | List of [cache export destinations](https://github.com/docker/buildx#--cache-tonametypetypekeyvalue) (eg. `type=local,dest=path/to/dir`) |
 | `secrets`           | List     | List of secrets to expose to the build (eg. `key=value`, `GIT_AUTH_TOKEN=mytoken`) |
+| `ssh`               | List     | List of SSH agent socket or keys to expose to the build (eg: `default|<id>[=<socket>|<key>[,<key>]]`) |
 
 ### outputs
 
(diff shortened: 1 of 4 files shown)


tonistiigi:
Not sure if something like ssh agent socket makes sense in GH actions context. Maybe this should just take direct ssh key value from secrets and create a temp file to expose via `--ssh`.

jesserockz:
That seems like it might be overcomplicating it and moving the action `ssh` key away from what the actual `--ssh` flag does. 
I start up an ssh agent with a key from a secret and forward that through with this method. 
The user would also be able to write the key to a file and expose that file themselves using this, if they choose to do it that way.

tonistiigi:
Sorry, I forgot this issue. I'm fine with this approach but looks like PR needs a rebase.

crazy-max (changes requested):
@jesserockz Need rebase like @tonistiigi said and you also need to run a `pre-checkin`. See https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md#submitting-a-pull-request

jesserockz:
Thanks guys. I will take a look at rebasing again on Monday as I merged master on Friday and it broke our deployments.

jesserockz:
I have rebased locally, but just waiting on a PR merge in our repo before I can push to the branch so it shows up here.

crazy-max (approved):
LGTM

tonistiigi (approved):

jesserockz:
Thanks for merging. Any chance of a new tag/release so I can update our workflows? @crazy-max @tonistiigi

crazy-max:
@jesserockz A new release will be created tomorrow.

PR #125, opened 2020-09-15
Improve example for OCI Image Spec annotations

itchyny:
Fill in title, description and url labels based on the repository information.
Use repository clone URL from `github.event` instead of an undocumented variable.
Fix licenses label to follow SPDX License Expressions.


crazy-max (approved):
LGTM thanks!

PR #134, opened 2020-09-21
Add missing checkout step to README 

jackton1:


crazy-max:
@jackton1 Checkout step is not required as the default context is Git in these workflows.

jackton1:
Strange but that's fine.

PR #136, opened 2020-09-22
Expose Git secret token if default context used

### LINKED ISSUE #135: git token secret should be exposed automatically when building from git
https://github.com/docker/build-push-action#git-context

> If you use this action in a private repository, you have to pass the GitHub Token as a secret named GIT_AUTH_TOKEN to be able to authenticate against it with buildx:

There shouldn't be a need to pass `GIT_AUTH_TOKEN` when building from git, or at least when building from the current repository. I believe GH gives access to this secret always and we should automatically forward it when building from the current repo. Even on public repos I think it is better to use token.

crazy-max:
Closes #135 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
             localhost:5000/name/app:latest
             localhost:5000/name/app:1.0.0
           secrets: |
-            GIT_AUTH_TOKEN=01234567890abcdefgh
+            GIT_AUTH_TOKEN=${{ github.token }}
       -
         name: Inspect
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,53 @@ on:
 jobs:
   git-context:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        buildx-version:
-          - latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2.3.2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
+      -
+        name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+          driver-opts: network=host
+      -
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -81,11 +78,10 @@ jobs:
       -
         name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
-  ```
-</details>
+```
 
-If you use this action in a private repository, you have to pass the [GitHub Token](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token)
-as a secret named `GIT_AUTH_TOKEN` to be able to authenticate against it with buildx:
+If you want to authenticate against a private repository, you have to use a secret named `GIT_AUTH_TOKEN` to be able
```
tonistiigi, line 83: "against another private repository" ?

tonistiigi: "building from current repository automatically uses the token and it does not need to be passed"

.github/workflows/ci.yml
```diff
@@ -50,7 +92,7 @@ jobs:
             localhost:5000/name/app:latest
             localhost:5000/name/app:1.0.0
           secrets: |
-            GIT_AUTH_TOKEN=${{ github.token }}
+            GIT_AUTH_TOKEN=01234567890abcdefgh
```
tonistiigi, line 95: not sure what this is testing or why the above build does not fail with an invalid token

crazy-max: I've missed that thanks. I think it passed because it's a public repo anyway.

tonistiigi: I think what is missing is that `githubToken` input should be ignored if `GIT_AUTH_TOKEN` is manually defined.

crazy-max: It is through [this cond](https://github.com/docker/build-push-action/pull/136/files#diff-a7af09f7fa8732629ff9579258b5584bR98-R100).

tonistiigi: hmm, maybe this is expected then from the server side. I wonder if you send a wrong auth header to the HTTP requests that does the clone of a public repository it still passes.

crazy-max: I will make a quick test locally

crazy-max: Ok so indeed the server seems to skip the header if the repository is public

tonistiigi (approved):

PR #157, opened 2020-10-03
Examples to push to multi-registries and cache to registry

### LINKED ISSUE #154: Push to multiple registries?
I would like to be able to push a built image (the same image) to multiple registries, e.g. Docker Hub _and_ Quay. Based on the current [usage examples](https://github.com/docker/build-push-action#usage), it seems like only one registry at a time would be possible. If this is already supported, some example or hint about how to achieve this would be nice in the usage examples.

crazy-max:
Closes #154 


metcalfc (approved):
LGTM

PR #177, opened 2020-10-13
Add example to upload image as an artifact

crazy-max:
> https://github.com/docker/build-push-action/actions/runs/305063606
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,6 +219,38 @@ jobs:
         if: always()
         uses: crazy-max/ghaction-dump-context@v1
 
+  upload-artifact:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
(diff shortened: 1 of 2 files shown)


tonistiigi:
Isn't it more common to upload the build result rather than the oci tarball of an image?

crazy-max:
@tonistiigi 

> Isn't it more common to upload the build result rather than the oci tarball of an image?

I use the OCI tarball output to be able to import the image to a local registry for testing purpose (could be used on PR event for example cause secrets cannot be shared to auth against a registry). Build result can only be exported through cache right?

tonistiigi:
Why wouldn't you then just push to the local registry. Or is this for loading the image in another job? Looks unclear why anyone would want to do this. We should explain what problem the user is trying to solve by uploading an artifact and then do the complete example.

crazy-max:
Seen together in Slack DM. Closing it as it could confuse some users.

PR #183, opened 2020-10-18
Simple readme update for dockerhub-description

syntaqx:
Was integrating into a new project and noticed the inconsistency with the project and the example usage provided here. Updated.


crazy-max (approved):
LGTM thanks

PR #185, opened 2020-10-19
Do not set --iidfile flag if local or tar exporters are used

crazy-max:
Can't use `--iidfile` if local or tar exporters are used. See https://github.com/docker/buildx/blob/a11cc8840e6c358e0fa3939dd80903e8a29327a0/build/build.go#L394

* Action failed: https://github.com/moby/buildkit/runs/1274435013?check_suite_focus=true#step:5:21
* Workflow: https://github.com/crazy-max/buildkit/blob/85f07f83db5c177724866bb2b925158eef748563/.github/workflows/release.yml#L137-L145

More tests have also be added for context module.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -93,15 +93,15 @@ async function getBuildArgs(inputs: Inputs, defaultContext: string, buildxVersio
   if (inputs.platforms.length > 0) {
     args.push('--platform', inputs.platforms.join(','));
   }
-  let isLocalOrTarExporter: boolean = false;
   await asyncForEach(inputs.outputs, async output => {
-    if (output.startsWith('type=local') || output.startsWith('type=tar')) {
-      isLocalOrTarExporter = true;
-    }
     args.push('--output', output);
   });
   // TODO: Remove platforms length cond when buildx >0.4.2 available on runner (docker/buildx#351)
-  if (inputs.platforms.length == 0 && !isLocalOrTarExporter && semver.satisfies(buildxVersion, '>=0.4.2')) {
+  if (
+    inputs.platforms.length == 0 &&
+    !buildx.isLocalOrTarExporter(inputs.outputs) &&
+    semver.satisfies(buildxVersion, '>=0.4.2')
+  ) {
     args.push('--iidfile', await buildx.getImageIDFile());
   }
   await asyncForEach(inputs.cacheFrom, async cacheFrom => {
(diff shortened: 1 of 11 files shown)


src/context.ts
```diff
@@ -83,12 +93,17 @@ async function getBuildArgs(inputs: Inputs, buildxVersion: string): Promise<Arra
   if (inputs.platforms.length > 0) {
     args.push('--platform', inputs.platforms.join(','));
   }
-  if (inputs.platforms.length == 0 || semver.satisfies(buildxVersion, '>=0.4.2')) {
-    args.push('--iidfile', await buildx.getImageIDFile());
-  }
+  let isLocalOrTarExporter: boolean = false;
   await asyncForEach(inputs.outputs, async output => {
+    if (output.startsWith('type=local') || output.startsWith('type=tar')) {
```
crazy-max, line 98: We check here if this is a local or tar exporter

tonistiigi: is there a better check we can do here that parses the csv? There is no requirement that `type=` needs to be the first element. Also, output default to `local` outputted path so that case should probably be checked as well.

crazy-max: > is there a better check we can do here that parses the csv? There is no requirement that type= needs to be the first element

Yes indeed, I've enhanced outputs parsing.

> Also, output default to `local` outputted path so that case should probably be checked as well.

I've created a [minimal job](https://github.com/docker/build-push-action/runs/1277374943?check_suite_focus=true#step:4:11) and it looks like `--iidfile` works when no output is specified.

tonistiigi: I'm being a bit too picky on this as far as practical cases are concerned but doing just `split(,)` gets confused with cases containing escapes and quotes. Is there a small library we can just plug in for this?

> and it looks like --iidfile works when no output is specified 

the case I had in mind was `-o .` (`-o <path>`) . I'd assume that also produces error. Although looking that empty value does not, maybe it should just produce an empty file or ignore it. But we can discuss that later separately.

crazy-max: > I'm being a bit too picky on this as far as practical cases are concerned but doing just split(,) gets confused with cases containing escapes and quotes. Is there a small library we can just plug in for this?

I've added the csv-parse lib to handle that and unit tests.

tonistiigi (approved):

PR #187, opened 2020-10-20
Ensure --progress=plain to avoid logs cutting

morlay:
When run `docker buildx build --progress=plain` will got all logs (default: "auto").

`build-push-action` used in github workflow. i think this always could be set.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -64,7 +64,7 @@ export async function getArgs(inputs: Inputs, buildxVersion: string): Promise<Ar
 }
 
 async function getBuildArgs(inputs: Inputs, buildxVersion: string): Promise<Array<string>> {
-  let args: Array<string> = ['build'];
+  let args: Array<string> = ['build', '--progress', 'plain'];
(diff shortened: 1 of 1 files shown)


crazy-max:
@morlay Progress plain mode is automatically used in GitHub Runners because no tty is allocated so I don't think this is necessary. Also please run a `pre-checkin` (see https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md#submitting-a-pull-request).

morlay:
@crazy-max i checked the results.

auto still cut logs of Dockerfile commands like RUN

morlay:
Sorry. 
It may buildx issues. when running in small screen.

Niek:
I'm also seeing a lot of repeat log output because of `--progress=auto` I assume. Is there any way to force plain progress?

tonistiigi:
`progress=auto` is not a separate mode. It is just a switch between `tty` and `plain` depending if build is performed from a terminal or not.

PR #186, opened 2020-10-20
Update readme typo "closed" -> "close"

NiteshOswal:

PR #189, opened 2020-10-20
docs(README): Improve example of `file` option so it's obvious that it is not relative to `context` option

jtomaszewski:
I just spent 30 min debugging "why is buildx not working on my github runner?", just to learn that the `file` option is not relative to the `context` option. (So basically if you want to build'n'push from a different directory, which has the `Dockerfile` and should be the `context`, you need to update both options.)

Especially since the error message that appears then isn't really helpful:

image

Changing from

```
          context: ./images/php/${{ env.PHP_VERSION }}/
```

to

```
          context: ./images/php/${{ env.PHP_VERSION }}/
          file: ./images/php/${{ env.PHP_VERSION }}/Dockerfile
```

fixed the problem in my case.

(The error message could be improved as well I guess, but let's say that's another nice-to-have thing.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,7 +658,7 @@ Following inputs can be used as `step.with` keys
 |---------------------|----------|------------------------------------|
 | `builder`           | String   | Builder instance (see [setup-buildx](https://github.com/docker/setup-buildx-action) action) |
 | `context`           | String   | Build's context is the set of files located in the specified [`PATH` or `URL`](https://docs.docker.com/engine/reference/commandline/build/) (default [Git context](#git-context)) |
-| `file`              | String   | Path to the Dockerfile (default `Dockerfile`) |
+| `file`              | String   | Path to the Dockerfile (default `./Dockerfile`) |
 | `build-args`        | List     | List of build-time variables |
 | `labels`            | List     | List of metadata for an image |
 | `tags`              | List/CSV | List of tags |


crazy-max:
@jtomaszewski Please sign your commit as explained in https://github.com/docker/build-push-action/pull/186#issuecomment-712984175

jtomaszewski:
@crazy-max done

README.md
```diff
@@ -153,8 +153,8 @@ jobs:
         name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: .
-          file: ./Dockerfile
+          context: ./images/app/
+          file: ./images/app/Dockerfile
```
crazy-max, line 157: Not necessary, please revert these changes.

README.md
```diff
@@ -658,7 +658,7 @@ Following inputs can be used as `step.with` keys
 |---------------------|----------|------------------------------------|
 | `builder`           | String   | Builder instance (see [setup-buildx](https://github.com/docker/setup-buildx-action) action) |
 | `context`           | String   | Build's context is the set of files located in the specified [`PATH` or `URL`](https://docs.docker.com/engine/reference/commandline/build/) (default [Git context](#git-context)) |
-| `file`              | String   | Path to the Dockerfile (default `Dockerfile`) |
+| `file`              | String   | Path to the Dockerfile (default `./Dockerfile`) |
```
crazy-max, line 661: LGTM for this one, thanks

crazy-max (changes requested):
@jtomaszewski 

> just to learn that the file option is not relative to the context option.

That's [the expected behavior](https://docs.docker.com/engine/reference/commandline/build/#extended-description) of the `build` command (with `docker buildx build` or `docker build`):

> The `docker build` command builds Docker images from a Dockerfile and a “context”. A build’s context is the set of files located in the specified `PATH` or `URL`. The build process can refer to any of the files in the context. For example, your build can use a COPY instruction to reference a file in the context.

README.md
jtomaszewski: Maybe... it used to be way different in `v1` ( `path` would apply to both ). Let's say notes in https://github.com/docker/build-push-action/blob/master/UPGRADE.md are enough though.

jtomaszewski:
Corrected @crazy-max

crazy-max (approved):
LGTM thanks!

Issue #140, opened 2020-09-26
v1 and v2 are very different: clarity required in Marketplace


shrink:
First, I'll preface this issue: I really like the improvements made to the action with the switch to buildx, I think @crazy-max has done great work, I'm now using v2 and intend to continue doing so! The re-focus of the action around being the _build_ step with delegation to other packages for things like login is great, it is a great step forward.

That said, I'm a little concerned that v2 is now the default. v1 of this action and v2 are _very_ different, the philosophy has changed: v1 does _everything_ you need and could be your entire workflow. That change alone wouldn't be of huge concern if it was clearly documented and a migration path was provided, but on top of changing the philosophy of the action the underlying technology has changed (`build` -> `buildx`). `buildx` is identified as "an experimental feature", "Tech Preview"[1] and "[...] must not be used in production environments"[2] but `docker/build-push-action` is _the_ GitHub action to use as part of a dockerised application's GitHub Workflow.

(My own experience: I've run into quite a few issues using buildx in a GitHub Workflow, which I've been able to resolve through investigation and experimentation, but it's definitely not yet ready _within the GitHub ecosystem_ as a drop-in replacement for build)

There are a few different use-cases for this action, but the ~two~ three that I think are most important when considering a change like this are:

1. Projects already using this action in their workflows -- there are [thousands](https://github.com/search?q=%22uses%3A+docker%2Fbuild-push-action%22&type=code) of projects that are using `@v1` but they'll be unable to find the documentation for it unless they understand the way to access old versions via the Marketplace
  i. bonus consideration: Lots of people learn about how to build their own Workflows by looking at examples from open source projects
2. Tutorials, guides and documentation across the internet referring to v1 and describing the behaviour that is no longer supported, e.g: [First Docker GitHub Action is here!](https://www.docker.com/blog/first-docker-github-action-is-here/)
3. Projects that cannot use the new functionality (`buildx`)

If you visit the marketplace page for v1 and compare it to the current marketplace page there is no recognition of the significant change.

https://github.com/marketplace/actions/build-and-push-docker-images?version=v1.0.0
https://github.com/marketplace/actions/build-and-push-docker-images

As I see it there are 2 options:

1. Modify the README on `master` so it makes very clear that there is an `@v1` and `@v2` that behave differently, that they're _effectively_ different actions and users need to make an active decision between which version they wish to use -- `@v2` is _not_ suitable for production use at companies where documented limitations must be respected.

[...]

PR #195, opened 2020-10-21
Throw error message instead of exit code

crazy-max:
Ref. #140 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,6 +197,64 @@ jobs:
         if: always()
         uses: crazy-max/ghaction-dump-context@v1
 
+  error:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


tonistiigi (commented):
Full stderr is probably too much. Eg. on build error whole progress is in `stderr`. The error should be at the end of the stderr.

crazy-max:
@tonistiigi  

> Full stderr is probably too much. Eg. on build error whole progress is in `stderr`. The error should be at the end of the stderr.

Done.

* https://github.com/docker/build-push-action/runs/1288686996?check_suite_focus=true#step:5:176
* https://github.com/docker/build-push-action/runs/1288686924?check_suite_focus=true#step:4:15

tonistiigi (approved):

PR #196, opened 2020-10-22
Check desired output status

crazy-max:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: ci
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
@@ -211,6 +212,7 @@ jobs:
         uses: docker/setup-buildx-action@v1
       -
         name: Build
+        id: docker_build
         continue-on-error: true
         uses: ./
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -219,6 +221,13 @@ jobs:
           platforms: linux/386,linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
           push: true
           tags: localhost:5000/name/app:latest
+      -
+        name: Check
+        run: |
+          if [ "${{ steps.docker_build.conclusion }}" != "success" ]; then
```
tonistiigi, line 227: don't quite understand this. This fails if `docker_build.conclusion` is not `success`. Meaning that although `docker_build` fails, because `continue-on-error: true` its conclusion is still success?

crazy-max: That could be disturbing but `conclusion` is the expected behavior of the step context. When a `continue-on-error` step fails, the `outcome` is `failure`, but the final `conclusion` is `success`. More info: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#steps-context

tonistiigi: yes, but it doesn't check anything then. As should the step succeed, `continue-on-error` will still leave the conclusion to success as well (afaics)

tonistiigi: Based on that link I think you also need to check that `outcome == "failure"` . Would be nice to also check for the actual error value matching the expected value, but don't think that is possible.

crazy-max: Yes indeed I should check the desired `outcome` too.

crazy-max: And yes unfortunately I can't spy on a step output.

tonistiigi (approved):

PR #206, opened 2020-10-27
Simplify "Complete workflow" with Docker meta action

crazy-max:
Use [Docker meta action](https://github.com/crazy-max/ghaction-docker-meta) to handle tags and labels based on GitHub actions events and Git metadata.

image

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -1,5 +1,4 @@
 # This workflow is provided just as an usage example and not for repo testing/verification
-# See https://github.com/docker/build-push-action#complete-workflow
 name: example
 
 on:
@@ -11,6 +10,9 @@ on:
     tags:
       - 'v*.*.*'
   pull_request:
+
+env:
+  DOCKER_IMAGE: localhost:5000/name/app
 
 jobs:
   docker:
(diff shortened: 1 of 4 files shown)


.github/workflows/example.yml
```diff
@@ -25,34 +27,15 @@ jobs:
...
-          fi
-          echo ::set-output name=version::${VERSION}
-          echo ::set-output name=tags::${TAGS}
-          echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+        name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: ${{ env.DOCKER_IMAGE }}
+          tag-sha: true
```
tonistiigi, line 35: I think it is hard to guess what exactly each of these options do by just looking at the example. Should at least add comments for every line. Or link to a separate example that uses this action for people who are ok with the learning curve to save some lines.

crazy-max: Done. I have simplified the workflow that should handle the most common cases.

tonistiigi: Comment all the options that are being shown (tag-edge etc) in upgrade. I'd also leave in an example of the old way as it is much flexible if there isn't an option that exactly matches. You can link to the right section in upgrade from the main readme.

crazy-max: Done

tonistiigi (approved):

Issue #208, opened 2020-10-27
Feedback for version 2


Brutus5000:
Hi guys,

this is not really an issue, but some feedback that you hopefully consider:
I'm aware I probably sound like a whiny crybaby, but please believe me that I admire your efforts... but seriously: with v2 you **raped** a good working Github action.

And yes yes, you're probably thinking: Why don't you fix it then yourself? But I'm already managing two dozen repositories for the [FAForever project](https://github.com/FAForever/) and I just need an easy and working CI infrastructure. [And v1 is still available, so technically no problem.]

**Why do I use this action?**
I prefer Github actions because they are easier to read and simplified building blocks over writing it in a shell script, even if they are a few (!) lines longer

And v1 of this action is the perfect solution for the workflow I need (with some Github if expression magic):
```
      - name: Build and push Docker images
        if: github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags')
        uses: docker/build-push-action@v1.1.1
        with:
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}
          repository: myproject/myrepo
          tag_with_ref: true
```

**What has changed?**
With v2 I need to split it off in 3 steps (+ an additional one considering #100):
- one step declaring the tag I want to push
- one step for a login (yes, if I want to push I ALWAYS want to login)
- one step for the build & push

Eventually I end up with 3x the lines of code that I had before. No thanks, I rather stick with the previous version or go back to manual shell commands.

**Here are some ideas to improve on v2:**
- I do understand that it makes sense to have a login action if you want to pull images from a protected repository. But why does this have to affect the build-and-push step? If already logged in for other reason, this is the special case not the default, so my proposal would be to either ignore this case (double login then) or to offer an login opt-out config flag instead.
- I do understand that the `tag_with_ref` flag was quirky and limits some scenarios like pushing multiple tags and I actually like that there is another way now to configure this now. But was it necessary to drop the existing flag? Maybe you can consider bringing it back under a more meaningful name.
- Looking at the upgrade guide it seems that `context` and `file` are mandatory now (didn't test it actually) but then again: why would I need to declare what is the 99% default case? So if it's not mandatory I'd to remove it from the upgrade document, and if it is mandatory I'd suggest to offer default settings for a classic Dockerfile build

I'd love to see a very powerful action as it is now, with sensible default settings preconfigured for the simple use cases.

Love you guys! Keep up the good work! :heart:

[...]

PR #212, opened 2020-11-03
Added dockerfile path to image label

Josh-01:
This change is part of an effort to improve traceability of resources which are built and deployed using GitHub workflows. We want to save the dockerfile path - which is used in building the image - through this action.
`Points to note:`
- Instead of the user specifying the path explicitly for this, we will use the path that the action uses to build the image - so no change for the user.
- This image label can later be used for traceability by any action that is used to deploy the image.

### DIFF
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -10,6 +10,7 @@ jest.spyOn(context, 'tmpDir').mockImplementation((): string => {
   const tmpDir = path.join('/tmp/.docker-build-push-jest').split(path.sep).join(path.posix.sep);
   if (!fs.existsSync(tmpDir)) {
     fs.mkdirSync(tmpDir, {recursive: true});
+    2;
   }
   return tmpDir;
 });
@@ -146,6 +147,8 @@ describe('getArgs', () => {
       [
         'buildx',
         'build',
+        '--label', 'org.opencontainers.image.source=https://github.com/docker/build-push-action.git',
+        '--label', 'dockerfile-path=./test/Dockerfile',
         '--platform', 'linux/amd64,linux/arm64',
         '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
         '--secret', 'id=GIT_AUTH_TOKEN,src=/tmp/.docker-build-push-jest/.tmpname-jest',
@@ -153,6 +156,59 @@ describe('getArgs', () => {
         '--builder', 'builder-git-context-2',
         '--push',
         'https://github.com/docker/build-push-action.git#heads/master'
+      ]
+    ],
+    [
+      '0.4.2',
+      new Map<string, string>([
+        ['context', '.'],
+        ['labels', 'org.opencontainers.image.source=UserProvidedSourceLabel'],
+        ['outputs', 'type=image,dest=./release-out']
+      ]),
+      [
+        'buildx',
+        'build',
+        '--label', 'org.opencontainers.image.source=UserProvidedSourceLabel',
+        '--label', 'dockerfile-path=Dockerfile',
+        '--output', 'type=image,dest=./release-out',
+        '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
+        '--file', 'Dockerfile',
+        '.'
+      ]
+    ],
+    [
+      '0.4.2',
+      new Map<string, string>([
+        ['context', '.'],
+        ['outputs', 'type=registry,dest=./release-out']
+      ]),
+      [
+        'buildx',
+        'build',
+        '--label', 'org.opencontainers.image.source=https://github.com/docker/build-push-action.git',
+        '--label', 'dockerfile-path=Dockerfile',
+        '--output', 'type=registry,dest=./release-out',
+        '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
+        '--file', 'Dockerfile',
+        '.'
+      ]
+    ],
+    [
+      '0.4.2',
+      new Map<string, string>([
+        ['context', '.'],
+        ['labels', 'org.opencontainers.image.source=UserProvidedSourceLabel'],
+        ['load', 'true']
+      ]),
+      [
+        'buildx',
+        'build',
+        '--label', 'org.opencontainers.image.source=UserProvidedSourceLabel',
+        '--label', 'dockerfile-path=Dockerfile',
+        '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
+        '--file', 'Dockerfile',
+        '--load',
+        '.'
       ]
     ]
   ])(
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -147,8 +147,31 @@ describe('getArgs', () => {
       [
         'buildx',
         'build',
-        '--label', 'org.opencontainers.image.source=https://github.com/docker/build-push-action.git',
-        '--label', 'dockerfile-path=./test/Dockerfile',
+        '--platform', 'linux/amd64,linux/arm64',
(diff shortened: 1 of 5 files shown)


crazy-max:
@Josh-01 This is already available through the [Docker meta action](https://github.com/crazy-max/ghaction-docker-meta#readme) that has been introduced in the [Handle tags and labels example](https://github.com/docker/build-push-action/#handle-tags-and-labels):



```yaml
jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v2
      -
        name: Docker meta
        id: docker_meta
        uses: crazy-max/ghaction-docker-meta@v1
        with:
          images: username/repo
      -
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
      -
        name: Build and push
        id: docker_build
        with:
          context: .
          file: ./Dockerfile
          platforms: linux/amd64,linux/arm64,linux/386
          tags: ${{ steps.docker_meta.outputs.tags }}
          labels: ${{ steps.docker_meta.outputs.labels }}
```

This will produce the [following labels](https://github.com/crazy-max/diun/runs/1341515728?check_suite_focus=true#step:3:31) for [crazy-max/diun](https://github.com/crazy-max/diun/) repo for example:

```
org.opencontainers.image.title=diun
org.opencontainers.image.description=Receive notifications when an image is updated on a Docker registry
org.opencontainers.image.url=https://github.com/crazy-max/diun
org.opencontainers.image.source=https://github.com/crazy-max/diun
org.opencontainers.image.version=edge
org.opencontainers.image.created=2020-11-02T10:32:44.842Z
org.opencontainers.image.revision=7ef743d66080926d00d7c1fbbffcc93feb83abbb
org.opencontainers.image.licenses=MIT
```

And will be handled through the GitHub package page to [display the README of the repo](https://github.com/users/crazy-max/packages/container/package/diun).

Josh-01:
Hi @crazy-max ,

I looked at the Docker-meta-action, and our use case is slightly different.
The Docker-meta-action enables the user to explicitly generate the standard OCI spec labels for an image and add any custom labels that they want to the image.

In our case , we want to : 
- add the path of the dockerfile ( that was used to build the image) to the image labels, 
- without the user having to configure anything extra as part of the build. 
Through this change:
Users who will use the docker-build-action to build their image,  will chose which dockerfile to use for the build, and then that choice is saved as a traceability field within the image itself, which can be later used by other actions. 

I have edited the labels in my change, into 'source (OCI spec label)' and a custom label 'dockerfilePath' to comply with the OCI specs.

crazy-max:
Hi @Josh-01,

> add the path of the dockerfile ( that was used to build the image) to the image labels,

This is not a standard annotation (as stated in [OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys)) and therefore does not seem appropriate to be included in this action. WDYT @tonistiigi?

> without the user having to configure anything extra as part of the build.

Sounds good to me to include `org.opencontainers.image.source` annotation as part of this action. We could also include the `revision` and `created` annotations aswell. But this must be done as part of the [context module](https://github.com/docker/build-push-action/blob/master/src/context.ts) and not the main one. You must also make sure that you do not overwrite user-defined labels through the action `labels` input. I can work with you on this PR and make the necessary modifications if you wish.

Josh-01:
Hi @crazy-max ,

Thank you for your inputs - I will make the following changes:
 - Move the changes to the context module from main.
 - Check to ensure user-added labels are not overwritten
 - add revision and created 

Is there a some way to include the dockerfile-path in the image labels - as that is one of the key traceability fields we require ? 
We should be able to add labels other than the ones specified in the OCI spec , as long as we don't use the OCI spec namespace.  Could we discuss adding the dockerfile-path label with a different or no namespace - please do let me know your thoughts on this. 

Thanks!

crazy-max:
@Josh-01 After looking around this feature, it turns out that injecting OCI annotations by default at the creation of the image leads to a confidentiality issue because some private repo are not intended to be disclosed even if the Docker image is public. I'll talk to @tonistiigi about it and we'll give you a feedback.

devdreamsolution (approved):

Josh-01:
Hi @crazy-max/ @tonistiigi ,

Here's a possible solution reg. the confidentiality issue: we could put an explicit flag input - 'traceData' (OFF by default, optional),
If it is specified as ON by user, only then the labels get added - else not.

We want users to be able use the action for building images and also get the traceability information as seamlessly as possible - which a simple on/off input would effectively be doing.
Please let us know your thoughts on this.

Thanks!

Josh-01:
Hi @crazy-max , @tonistiigi , 
Is there any update on this ? ... please let me know if we need to discuss this further.

ammohant:
> Hi @crazy-max / @tonistiigi ,
> 
> Here's a possible solution reg. the confidentiality issue: we could put an explicit flag input - 'traceData' (FALSE by default, optional),
> If it is specified as TRUE by user, only then the labels get added - else not.
> 
> We want users to be able use the action for building images and also get the traceability information as seamlessly as possible - which a simple on/off input would effectively be doing.
> Please let us know your thoughts on this.
> 
> Thanks!

Hi  @crazy-max / @tonistiigi  - What is your view on user providing consent as part of the explicit input provided ? Pls. let us know if you see any issues with this approach.

ammohant:
Hi @crazy-max / @tonistiigi   - Can you pls let us know if  having an explicit flag as input(default  NO) helps mitigate the confidentiality concerns ?

crazy-max:
Sorry for the huge delay on this PR but after consideration it seems more appropriate for this to be handled upstream. I think the OCI `org.opencontainers.image.source` annotation is sufficient but it should not be an opt-out model for the user. One could also consider creating a new OCI annotation for Dockerfile tracability like `org.opencontainers.image.definition` (suggest to open an issue on [image-spec](https://github.com/opencontainers/image-spec) repo).

One solution would be to use our [metadata action](https://github.com/docker/metadata-action) which you could embed into a [workflow template](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template)  if you want to make things as generic as possible.

As I discussed with @clarkbw I think it's better to keep the actions in the current state (login + metadata + qemu + buildx + build-push) and merge them into a single step when nested composite actions will be available. But would need to focus on https://github.com/actions/runner/issues/646 like I said in https://github.com/docker/build-push-action/issues/208#issuecomment-729868550.

clarkbw:
> One could also consider creating a new OCI annotation for Dockerfile tracability like `org.opencontainers.image.definition` (suggest to open an issue on [image-spec](https://github.com/opencontainers/image-spec) repo).

Agree with this. It would be good to get this discussion going with the OCI group.

For now you could append this as a tag to your existing workflows and still retain the OCI labels like so:

```yaml
jobs:
  # Push image to GitHub Packages.
  # See also https://docs.docker.com/docker-hub/builds/
  push:

    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2

      - id: meta
        uses: docker/metadata-action@v3
        with:
          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}

      - name: Login to ${{ env.REGISTRY }}
        uses: docker/login-action@v1 
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.actor }}
          password: ${{ github.token }}

      - name: Build and push
        with:
          tags: ${{ steps.meta.outputs.tags }}
          labels: |
            ${{ steps.meta.outputs.labels }}
            
            docker-file=https://github.com/github/container-registry-playground.git#Dockerfile
```

Perhaps you could use a script and output to determine the location of the `Dockerfile`.

PR #221, opened 2020-11-08
Set container label "org.opencontainers.image.source" to establish a repository connection on github

andygrunwald:
In the Upgrade Path, the container label `org.opencontainers.image.source` gets set:

``` 
org.opencontainers.image.source = "https://github.com/<OWNER>/<NAME>.git"
```

When using the Github Container Registry, there is an option to [Connecting a repository to a container image](https://docs.github.com/en/free-pro-team@latest/packages/managing-container-images-with-github-container-registry/connecting-a-repository-to-a-container-image).

GitHub itself can also create a connection automatically, depending on the value of the label `org.opencontainers.image.source`.
With the `.git`-prefix this connection can't be established:

Screen Shot 2020-11-08 at 19 47 03

Screen Shot 2020-11-08 at 19 47 11

This PR switches the label to a value without the `.git` prefix:

``` 
org.opencontainers.image.source = "https://github.com/<OWNER>/<NAME>"
```

With this, the connection can be established:

Screen Shot 2020-11-08 at 20 10 53

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -140,7 +140,7 @@ steps:
       push: ${{ github.event_name != 'pull_request' }}
       tags: ${{ steps.prep.outputs.tags }}
       labels: |
-        org.opencontainers.image.source=https://github.com/${{ github.repository }}
+        org.opencontainers.image.source=${{ github.event.repository.html_url }}
         org.opencontainers.image.created=${{ steps.prep.outputs.created }}
         org.opencontainers.image.revision=${{ github.sha }}
 ```
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -140,7 +140,7 @@ steps:
       push: ${{ github.event_name != 'pull_request' }}
       tags: ${{ steps.prep.outputs.tags }}
       labels: |
-        org.opencontainers.image.source=${{ github.event.repository.clone_url }}
+        org.opencontainers.image.source=https://github.com/${{ github.repository }}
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -140,7 +140,7 @@ steps:
       push: ${{ github.event_name != 'pull_request' }}
       tags: ${{ steps.prep.outputs.tags }}
       labels: |
-        org.opencontainers.image.source=${{ github.event.repository.clone_url }}
+        org.opencontainers.image.source=https://github.com/${{ github.repository }}
```
crazy-max, line 143:
```suggestion
        org.opencontainers.image.source=${{ github.event.repository.html_url }}
```

Would be better to use event object for that.

crazy-max (changes requested):
@andygrunwald Yes indeed [I fix that on Docker meta action](https://github.com/crazy-max/ghaction-docker-meta/commit/8d7fa04f07e6cbdd6d1efb7e3a3e56d7da302cc5) but forgot in the UPGRADE notes. Thanks for your input.

UPGRADE.md
andygrunwald: Good one!

crazy-max (changes requested):
LGTM but sign your commit: `git commit --amend -s --no-edit`

andygrunwald:
@crazy-max Thanks. Done. Signed.

crazy-max (approved):

PR #244, opened 2020-12-02
Use default behavior for file flag

liboz:
Our team hit this issue today, but with `buildx` the file argument needs to be specified from the root directory not just the relative directory from the `context` (`PATH`) argument .

See:
image

This updates the readme to reflect this fact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,7 +608,7 @@ Following inputs can be used as `step.with` keys
 |---------------------|----------|------------------------------------|
 | `builder`           | String   | Builder instance (see [setup-buildx](https://github.com/docker/setup-buildx-action) action) |
 | `context`           | String   | Build's context is the set of files located in the specified [`PATH` or `URL`](https://docs.docker.com/engine/reference/commandline/build/) (default [Git context](#git-context)) |
-| `file`              | String   | Path to the Dockerfile (default `./Dockerfile`) |
+| `file`              | String   | Path to the Dockerfile. (default `{context}/Dockerfile`) |
 | `build-args`        | List     | List of build-time variables |
 | `labels`            | List     | List of metadata for an image |
 | `tags`              | List/CSV | List of tags |
--- a/src/context.ts
+++ b/src/context.ts
@@ -57,7 +57,7 @@ export function tmpNameSync(options?: tmp.TmpNameOptions): string {
 export async function getInputs(defaultContext: string): Promise<Inputs> {
   return {
     context: core.getInput('context') || defaultContext,
-    file: core.getInput('file') || 'Dockerfile',
+    file: core.getInput('file'),
     buildArgs: await getInputList('build-args', true),
     labels: await getInputList('labels', true),
     tags: await getInputList('tags'),
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -147,7 +147,6 @@ describe('getArgs', () => {
         'buildx',
         'build',
         '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
-        '--file', 'Dockerfile',
         '.'
       ]
     ],
@@ -162,7 +161,6 @@ describe('getArgs', () => {
         '--build-arg', 'MY_ARG=val1,val2,val3',
         '--build-arg', 'ARG=val',
         '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
-        '--file', 'Dockerfile',
         'https://github.com/docker/build-push-action.git#test-jest'
       ]
     ],
@@ -177,7 +175,6 @@ describe('getArgs', () => {
         '--tag', 'name/app:7.4',
         '--tag', 'name/app:latest',
         '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
-        '--file', 'Dockerfile',
         'https://github.com/docker/build-push-action.git#test-jest'
       ]
     ],
@@ -194,7 +191,6 @@ describe('getArgs', () => {
         '--label', 'org.opencontainers.image.title=buildkit',
         '--label', 'org.opencontainers.image.description=concurrent, cache-efficient, and Dockerfile-agnostic builder toolkit',
         '--output', 'type=local,dest=./release-out',
-        '--file', 'Dockerfile',
         '.'
       ]
     ],
@@ -208,7 +204,6 @@ describe('getArgs', () => {
         'buildx',
         'build',
         '--platform', 'linux/amd64,linux/arm64',
-        '--file', 'Dockerfile',
         '.'
       ]
     ],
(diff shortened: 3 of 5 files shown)


crazy-max:
@liboz 

> but with buildx the file argument needs to be specified from the root directory not just the relative directory from the context (PATH) argument .

This is the same behavior as `docker build`. See https://github.com/docker/build-push-action/pull/189#pullrequestreview-516680807

liboz:
@crazy-max You're right and I suppose that was mostly some inexperience on our part. I think updating the documents to reflect the fact that the implicit default path includes the `context` would be a helpful change for everyone involved though.

muru:
@liboz and @crazy-max Is the default really `{context}/Dockerfile`? That's what `docker build`'s behaviour is, but not the `docker/build-push-action@v2` action's. If I use `context: foo` without a `file` argument, it still looks for the `Dockerfile` in the current working directory. If I specify both `context: foo` and `file: foo/Dockerfile`, OTOH, the action works. Possibly because of [this line](https://github.com/docker/build-push-action/blob/v2.2.1/src/context.ts#L60):
```ts
file: core.getInput('file') || 'Dockerfile',
```
So when specifying the the path context, the file has to be set too. Is that intended? It seems a bit redundant. Maybe that line should be something like (pardon my TypeScript):
```ts
file: core.getInput('file') || [core.getInput('context'), "Dockerfile"].filter(Boolean).join("/"),
```

liboz:
@muru After doing some more poking around, I think you're right and that suggestion makes sense.

crazy-max:
@muru @liboz As I said, this is [the expected and same behavior](https://docs.docker.com/engine/reference/commandline/build/#extended-description) for `docker build`, `docker buildx build` and `build-push-action`:

> The `docker build` command builds Docker images from a Dockerfile and a “context”. A build’s context is the set of files located in the specified `PATH` or `URL`. The build process can refer to any of the files in the context. For example, your build can use a COPY instruction to reference a file in the context.

You can test it yourself locally if you want:

Given folder structure and files:

```
project
│   Dockerfile    
│
└───test
│   │   Dockerfile.test
│   │   app.txt
```

```
# ./Dockerfile
FROM alpine
RUN echo "Hello world!"
```

```
# ./test/Dockerfile.test
FROM alpine
RUN echo "Hello world!"
COPY app.txt ./
```

And the following commands and expected result:

```shell
$ docker build .
OK

$ docker build -f Dockerfile.test test
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount199981535/Dockerfile.test: no such file or directory

$ docker build -f ./test/Dockerfile.test ./test
OK

$ docker build -f ./test/Dockerfile.test .
failed to compute cache key: "/app.txt" not found: not found
```

muru:
@crazy-max that's all fine, I'm not disputing that. What I'm talking of is this case:

```
project   
└───test
│   │   Dockerfile
│   │   app.txt
```
Here, `docker build test` will use `test/Dockerfile` without having to specify it using `-f`. That's not the case with the action, however.

crazy-max:
@muru 

> Here, docker build test will use test/Dockerfile without having to specify it using -f. That's not the case with the action, however.

Yes if you don't specify a Dockerfile that's right

crazy-max:
@muru After what we have just said we could replace this line with:

https://github.com/docker/build-push-action/blob/0db984c1826869dcd0740ff26ff75ff543238fd9/src/context.ts#L60

With:

```ts
 file: core.getInput('file'),
```

And remove:

https://github.com/docker/build-push-action/blob/0db984c1826869dcd0740ff26ff75ff543238fd9/action.yml#L19

And rely on default behavior.

README.md
```diff
@@ -608,7 +608,7 @@ Following inputs can be used as `step.with` keys
 |---------------------|----------|------------------------------------|
 | `builder`           | String   | Builder instance (see [setup-buildx](https://github.com/docker/setup-buildx-action) action) |
 | `context`           | String   | Build's context is the set of files located in the specified [`PATH` or `URL`](https://docs.docker.com/engine/reference/commandline/build/) (default [Git context](#git-context)) |
-| `file`              | String   | Path to the Dockerfile (default `./Dockerfile`) |
+| `file`              | String   | Path to the Dockerfile. If changed from the default, the full path must be specified from the working directory (default `{context}/Dockerfile`) |
```
crazy-max, line 611: Should be enough with the default behavior now:

```suggestion
| `file`              | String   | Path to the Dockerfile (default `PATH/Dockerfile`) |
```

src/context.ts
```diff
@@ -14,7 +14,7 @@ let _defaultContext, _tmpDir: string;
 
 export interface Inputs {
   context: string;
-  file: string;
+  file?: string;
```
crazy-max, line 17: Keep as required:

```suggestion
  file: string;
```

src/context.ts
```diff
@@ -57,7 +57,7 @@ export function tmpNameSync(options?: tmp.TmpNameOptions): string {
 export async function getInputs(defaultContext: string): Promise<Inputs> {
   return {
     context: core.getInput('context') || defaultContext,
-    file: core.getInput('file') || 'Dockerfile',
+    file: core.getInput('file') || undefined,
```
crazy-max, line 60:
```suggestion
    file: core.getInput('file'),
```

crazy-max (changes requested):
Also update `action.yml`. See my comment https://github.com/docker/build-push-action/pull/244#issuecomment-744579942

README.md
liboz: I think it's still useful to have the default of (
```
default `{context}/Dockerfile`)
```
shown as that is what the command line defaults to.

liboz:
@crazy-max I tried to fix all the tests. However, when I run it locally I only find/execute 44 tests while in the CI, it seems to be executing 45 tests (with arguments that don't seem to be in the test file at all). Do you have any idea where that test might be coming from?

crazy-max (changes requested):
@liboz You forgot to remove [this line](https://github.com/docker/build-push-action/blob/181ed59064b7c509e0b1252a36c4d3faa278f6bb/__tests__/context.test.ts#L253)

crazy-max:
@liboz Can you squash your commits and rebase please?

liboz:
@crazy-max  Done

crazy-max:
@liboz About the failed test, you forgot to remove [this line](https://github.com/liboz/build-push-action/blob/d5702a3701e0b03092b64ea03c8e31e47caf9783/__tests__/context.test.ts#L178).

liboz:
@crazy-max Ah, thanks for pointing that out. It was because I hadn't merged master. Thanks for reviewing 😃

crazy-max (approved):
PTAL @tonistiigi

tonistiigi (approved):

PR #296, opened 2021-02-14
Allow to use secret file mount

### LINKED ISSUE #293: Support secret file mount
The action currently supports `secrets` which are string values that the action writes to temporary files behind the scenes. It would be nice to be able to mount a secret file that already exists on disk, passing the arguments through to the builder's `--secret` option.

For example, say I construct a `.npmrc` file earlier in the workflow and want to mount it as a secret in order to pull from a private registry in the Dockerfile:

```yaml
env:
  NPM_REGISTRY: myorg.jfrog.io/myorg/api/npm/npm/

jobs:
  build:
    # ...
    steps:
      # ...

      # This could also be encapsulated as an action. Either way, a .npmrc file is generated.
      - name: NPM auth
        run: |
          cat <<EOF > "$HOME/.npmrc"
          @myorg:registry=https://$NPM_REGISTRY
          //$NPM_REGISTRY:_password=$(echo -n "${{ secrets.NPM_PASSWORD }}" | openssl base64)
          //$NPM_REGISTRY:username=${{ secrets.NPM_USER }}
          //$NPM_REGISTRY:email=${{ secrets.NPM_EMAIL }}
          //$NPM_REGISTRY:always-auth=true
          EOF

      - name: Install NPM dependencies
        run: npm ci

      # ... Testing and checks using the installed node modules

      - name: Docker build
        with:
          context: .
          secrets: npmrc=$HOME/.npmrc  # <-- This writes the string "$HOME/.npmrc" as file contents. Not what we want.
```

With the current version of this action, I have to either read in the file and store it as env/output in a previous step or store the full contents of the file as a GitHub secret:

```yaml
secrets: |
  "npmrc=${{ secrets.JFROG_NPMRC }}"
```

Storing the different variables and dynamically constructing the file is more convenient because I don't have to regenerate the file every time the credentials are rotated.

Secret file mounts could be supported with a new input:

```yaml
secret-files: npmrc=$HOME/.npmrc
```

or a special syntax in the `secrets` input:

``` yaml
secrets: file:npmrc=$HOME/.npmrc
```

crazy-max:
Fixes #293 

@ngraef You can try with `crazy-max/docker-build-push-action@secret-file`.


ngraef:
Works great! Thanks!

tonistiigi:
I'm a bit worried about the ambiguity here. Would be very surprising if a secret value would ever collide with a path. Could we somehow make it clear that the user expects the value to be a path?

ngraef:
@tonistiigi My suggestions in #293 were a new action input (`secret-files: npmrc=path/to/file`) or a special syntax in the `secrets` input that could be intermixed with string secrets (`secrets: file:npmrc=path/to/file`). What do you think?

tonistiigi:
@ngraef SGTM, `secrets: file:npmrc=path/to/file` is still at least theoretically ambiguous as `:` could be part of the ID as well.

crazy-max:
@tonistiigi @ngraef 

> I'm a bit worried about the ambiguity here. Would be very surprising if a secret value would ever collide with a path. Could we somehow make it clear that the user expects the value to be a path?

A new input SGTM.

crazy-max:
@tonistiigi @ngraef `secret-files` input has been added. Let me know if it looks good to you (can be tested with `crazy-max/docker-build-push-action@secret-file`).

ngraef:
Just tested. `secret-files` works for me.

Issue #252, opened 2020-12-14
Pretty big cache


MarcelCoding:
### Troubleshooting
The cache can grow with big images pretty fast, because old entries aren't deleted.

### Configuration

* Repository URL (if public): https://github.com/MarcelCoding/docker-desktop-vnc/blob/02fb4a97d5a75e40c467b856538efa443d1077cc/.github/workflows/docker-publish.yml#L28-L60
* Build URL (if public): https://github.com/MarcelCoding/docker-desktop-vnc/runs/1549446282?check_suite_focus=true#logs

```yml
[...]

      - name: Cache Docker layers
        uses: actions/cache@v2
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-

[...]

      - name: Build
        uses: docker/build-push-action@v2
        with:
          push: false
          tags: ${{ steps.prepare.outputs.image }}
          platforms: ${{ env.DOCKER_PLATFORMS }}
          cache-from: type=local,src=/tmp/.buildx-cache
          cache-to: type=local,dest=/tmp/.buildx-cache
          context: .

[...]
```

### Logs
[logs_37.zip](https://github.com/docker/build-push-action/files/5688147/logs_37.zip)

### My soloution
Add a configuration option `clean-cache`, witch executes he following command `docker system prune -f --filter "until=5h"`.

crazy-max:
@MarcelCoding 

> The cache can grow very quickly with large images, since old entries are not deleted.

Yes you're right atm caches are copied over the existing cache so it's keep growing. Can you open an issue on [buildkit repo](https://github.com/moby/buildkit) about that please? In the meantime you can do this:

```yaml
[...]

      - name: Cache Docker layers
        uses: actions/cache@v2
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-

[...]

      - name: Build
        uses: docker/build-push-action@v2
        with:
          push: false
          tags: ${{ steps.prepare.outputs.image }}
          platforms: ${{ env.DOCKER_PLATFORMS }}
          cache-from: type=local,src=/tmp/.buildx-cache
          cache-to: type=local,dest=/tmp/.buildx-cache-new
          context: .

[...]

      - name: Move cache
        run:
          rm -rf /tmp/.buildx-cache
[...]

PR #299, opened 2021-02-16
Enhance documentation

### LINKED ISSUE #215: Docs: needs more clarifications on default behaviour
The README currently describes the the two different usage of [Git context](https://github.com/docker/build-push-action#git-context) and [Path context](https://github.com/docker/build-push-action#path-context)..

however it doesn't **clearly / explicitly** explain that the **default behavior is to checkout the repository at the current github ref** _(based on the event that triggered the action)_ and use that to build the image.

meaing that **any file mutation / creation / updates to the repo in steps that preceed the build step are completely lost and ignored by the build action** despite using the `github/checkout` action in previous steps.

I'd argue this is **not** a clear and obvious default for the average user who at first glance by looking at the README file, AND given the reset of the GitHub Actions ecosystem defaults ... 

if changing the default behaviour is too big of a change, then at least consider updating the docs to explicitly state this and indicate the need to set the `context` and `file` arguments as well as the `github/checkout` step.

I would even include examples in the README of content modification in the repo before the build step.

thanks

crazy-max:
Fixes #215 

And also:

* Split documentation to a docs directory
* Add note about issue with local cache backend (keeps growing)
* Update GitHub cache workflow with [temp fix](https://github.com/docker/build-push-action/issues/252#issuecomment-744400434)
* Note about future GitHub cache backend
* Workflow example to use secrets

cc @tonistiigi @MarcelCoding @ahmadnassri

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,18 @@
 
 ## Upgrade from v1
 
-`v2` of this action includes significant updates and now uses Docker [Buildx](https://github.com/docker/buildx). It
-works with 3 new actions ([login](https://github.com/docker/login-action), [setup-buildx](https://github.com/docker/setup-buildx-action)
-and [setup-qemu](https://github.com/docker/setup-qemu-action)) that we have created. It's also rewritten as a
-[typescript-action](https://github.com/actions/typescript-action/) to be as close as possible of the
-[GitHub Runner](https://github.com/actions/virtual-environments) during its execution.
+`v2` of this action includes significant updates and now uses Docker [Buildx](https://github.com/docker/buildx). It's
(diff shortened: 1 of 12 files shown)


ahmadnassri:
looks good, and a good description for the topic in #215

tonistiigi:
After the first example can we had a section explaining setup-* are optional and when they are needed. And maybe a link to a longer writeup in the advanced section explaining usage of driver and the main differences(multi-platform, load, cache-export).

crazy-max:
@tonistiigi Thanks for your feedback. I will make some changes based on this.

tonistiigi (approved):

PR #338, opened 2021-04-06
Add network input

### LINKED ISSUE #317: support for --network option when build
I'm not seeing the ability to specify --network option? Is this supported?

crazy-max:
Fixes #317 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,6 +290,30 @@ jobs:
         if: always()
         uses: crazy-max/ghaction-dump-context@v1
 
+  network:
+    runs-on: ubuntu-latest
+    steps:
+      -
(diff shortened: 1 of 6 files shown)


tonistiigi (approved):
I wonder if we should not promote this. We can still add it for backward compatibility. Usually, this is the wrong pattern to use. User should do `RUN --network=host` in Dockerfile and set entitlement instead.

crazy-max:
@tonistiigi 

> I wonder if we should not promote this. We can still add it for backward compatibility. Usually, this is the wrong pattern to use. User should do `RUN --network=host` in Dockerfile and set entitlement instead.

Yes, indeed, it would make more sense to do it that way.

akay222 (commented):
Mookie

PR #347, opened 2021-04-21
Fix ref spec for default Git context

### LINKED ISSUE #336: Git context always checks out the branch tip
### Behaviour

Using Git context with the default settings, build-push-action always checks out the branch tip regardless the commit for which the workflow run is triggered.
This causes confusing Actions results when multiple commits are pushed successively to the same branch or when Actions workflow for some old commit is triggered to re-run.

Expected fix is to make build-push-action check out $GITHUB_SHA instead of $GITHUB_REF.

#### Steps to reproduce this issue

1. Push a commit with a Dockerfile and the following GitHub actions job definition.
2. Push another commit with a modified Dockerfile to the same branch.
3. Rerun the Actions workflow for the first commit.

#### Expected behaviour

Step 3 should build image using Dockerfile from Step 1.

#### Actual behaviour

Step 3 uses Dockerfile from Step 2.

### Configuration

* Repository URL (if public): https://github.com/hanazuki/build-push-action-test
* Build URL (if public): https://github.com/hanazuki/build-push-action-test/actions/runs/721615893

This repository has two commits: the [good commit](https://github.com/hanazuki/build-push-action-test/commit/6cdb58acd28c0ec14f5f365ba0789932c674abe0) and [bad commit](https://github.com/hanazuki/build-push-action-test/commit/637e0c54cee1fb6b7b0294732fd88ffc86829596). The bad commit is now the tip of the master branch.
The good commit has a Dockerfile that builds successfully while the bad commit has a Dockerfile that fails to build.

During the Actions job for the good commit, restarted after the bad commit is pushed to the branch, build-push-action checks out the bad commit. For that reason, CI for good commit fails.

```yml
name: Build
on: push
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: docker/setup-buildx-action@v1
```

### Logs

[logs_3.zip](https://github.com/docker/build-push-action/files/6262732/logs_3.zip)

crazy-max:
Fixes #336 

cc. @hanazuki


hanazuki:
@crazy-max Thank you for the implementation!
I tested it with the `push` and `schedule` events, and it worked as expected: it checks out GITHUB_SHA, for which the run is triggered instead of the tip of GITHU_REF.

For the `pull_request`, as intended by the code, the action checks out the merge commit of the base branch and the topic branch.
To explain this case to other users, let us assume we have a pull request that merges `topic` into `master`, and two commits A and B are pushed to `topic` (as described in the diagram).
```
   (master)
      v
X <-- Y
       \
        \
         A <-- B
               ^
            (topic)
```
A `pull_request` event is fired for the merge commit of the base branch and each commit pushed to the topic branch. In this case, they are `A+Y` and `B+Y`. But this implementation is designed to always check out the merge commit of the base branch and the tip of the topic branch, i.e., `B+Y`.

I suppose @crazy-max designed so because `docker buildx build` does not support checking out `A+Y` by specifying its commit SHA1. Since `actions/checkout@v2` can check out `A+Y` directly, perhaps we can later relax this restriction by improving buildx, buildkit, or some other component (But I didn't investigate them further yet).

So overall, I think this patch implements the current best possible approach.

crazy-max:
@hanazuki 

> I suppose @crazy-max designed so because `docker buildx build` does not support checking out `A+Y` by specifying its commit SHA1.

Yes indeed it's not possible atm to specify the commit because the current parser does not support it. [Git URLs accept the context configuration](https://docs.docker.com/engine/reference/commandline/build/#git-repositories) in their fragment section, separated by a colon (`:`) which in our case would be interpreted as a subdir (not a commit ID) and therefore would failed:

```
docker buildx build https://github.com/docker/rootfs.git#refs/pull/6/merge:6adaa44ee2074ea02066d5a9625394722131a1f3
```

Same applies for [classic `docker build`](https://github.com/moby/moby/blob/e0170da0dc6e660594f98bc66e7a98ce9c2abb46/builder/remotecontext/git/gitutils.go#L112-L122).

cc. @tonistiigi @thaJeztah

thaJeztah:
@crazy-max is this a bug in how buildkit gets the reference? looking at the [`docker build` docs](https://docs.docker.com/engine/reference/commandline/build/#git-repositories), specifying `myrepo.git#mybranch`, should be equivalent to `refs/heads/mybranch`, correct?

> Yes indeed it's not possible atm to specify the commit because the current parser does not support it

So it _is_ possible to specify a commit, but I guess in this case the commit is not in the main repository? Otherwise
`myrepo.git#<commit>` should work

crazy-max:
@thaJeztah 

Not a bug actually but I think the parser could be enhanced to allow more complex refspec. Could be useful in our case to be able to check out `A+Y` as @hanazuki demonstrates.

hanazuki:
I think that is not a bug or a lack of feature in the context URL parser, but a corner case in how the builder checks out the given reference

Here is a pull request with two commits pushed on top of the master branch: <https://github.com/hanazuki/build-push-action-test2/pull/1>. The commits in my previous diagram correspond to these SHA1's in this case:
```
  A = f816db91ffa3a4a93829616040173ad8478ffc74 https://github.com/hanazuki/build-push-action-test2/runs/2427348592
A+Y = 02992a9474f4ce08274eb849428d4877bc45caca https://github.com/hanazuki/build-push-action-test2/runs/2427378367
  B = fac0dda8fdf991c6f12fa7b409e0773d27310c31 https://github.com/hanazuki/build-push-action-test2/runs/2427354980
B+Y = 8435bb2233da72790c6379702b3376e7b43a341b https://github.com/hanazuki/build-push-action-test2/runs/2427355087
```

For B+Y, both of the following commands succeed:
```
# git fetch
git init testBY
cd testBY
git remote add origin https://github.com/hanazuki/build-push-action-test2.git
git fetch origin 8435bb2233da72790c6379702b3376e7b43a341b
git checkout FETCH_HEAD

# docker build
docker buildx build https://github.com/hanazuki/build-push-action-test2.git#8435bb2233da72790c6379702b3376e7b43a341b
```

But for A+Y, `docker build` fails while `git fetch` succeeds:
```
# git fetch
git init testAY
cd testAY
git remote add origin https://github.com/hanazuki/build-push-action-test2.git
git fetch origin 02992a9474f4ce08274eb849428d4877bc45caca
git checkout FETCH_HEAD

# docker build
docker buildx build https://github.com/hanazuki/build-push-action-test2.git#02992a9474f4ce08274eb849428d4877bc45caca
```

hanazuki:
Just noticed that plain `docker build` can check out A+Y by SHA1 in this case.
```
docker build https://github.com/hanazuki/build-push-action-test2.git#8435bb2233da72790c6379702b3376e7b43a341b
docker build https://github.com/hanazuki/build-push-action-test2.git#02992a9474f4ce08274eb849428d4877bc45caca
```

hanazuki:
For record, my testing environment is a Debian sid box with the following pieces of software:
<details>

```
10:32 kasumi@dev% docker version
Client:
 Version:           20.10.5+dfsg1
 API version:       1.41
 Go version:        go1.15.9
 Git commit:        55c4c88
 Built:             Thu Apr  8 23:54:54 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.5+dfsg1
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.15.8
  Git commit:       363e9a8
  Built:            Tue Mar  9 16:53:21 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4~ds1
  GitCommit:        1.4.4~ds1-2
 runc:
  Version:          1.0.0~rc93+ds1
  GitCommit:        1.0.0~rc93+ds1-3
 docker-init:
  Version:          0.19.0
  GitCommit:
10:32 kasumi@dev% docker buildx version
github.com/docker/buildx v0.5.1 11057da37336192bfc57d81e02359ba7ba848e4a
10:32 kasumi@dev% docker buildx inspect default
Name:   default
Driver: docker

Nodes:
Name:      default
Endpoint:  default
Status:    running
Platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/386, linux/arm/v7, linux/arm/v6, linux/s390x
10:32 kasumi@dev% git --version
git version 2.31.1
```

</details>

crazy-max:
@hanazuki 

> I think that is not a bug or a lack of feature in the context URL parser, but a corner case in how the builder checks out the given reference

Yes that's right, good point!

> Just noticed that plain (non-buildkit) `docker build` can check out A+Y by SHA1 in this case.

Just tested myself and it works for me too.

crazy-max:
@hanazuki 

> Since `actions/checkout@v2` can check out `A+Y` directly, perhaps we can later relax this restriction by improving buildx, buildkit, or some other component (But I didn't investigate them further yet).

This might be linked to [this line in buildkit](https://github.com/moby/buildkit/blob/78b7ca9f36257ba7fc9ef4c339c431a2d7998718/source/git/gitsource.go#L499). WDYT @tonistiigi?

PR #373, opened 2021-05-28
Set mode=max to cache intermediate stages (docs update only)

saiya:
Add `mode=max` option in the cache.md documentation.

Background: buildx caches layers only in the final stage by default.

To speedup multi-stage Dockerfile build, better to set `mode=max` in CI: https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to


crazy-max:
@saiya Fixed through #406 (fyi `inline` cache does not support `max` mode)

saiya:
Wonderful! Thank you.

Issue #182, opened 2020-10-18
Inconsistent dockerignore behavior from other builders


syntaqx:
> ⚠️ I'm not entirely sure if this is an issue with GitHub Actions, or this action specifically, but I didn't know where else to report it so figured this would be a good start. It's also quite possible that I'm an idiot and I'm doing something wrong that hopefully will be obvious to someone else, but I'm running into a wall with this, so here goes nothing: __#insecurities__

### Behaviour

It appears to me that the `.dockerignore` is being completely ignored when building using this action, which is an obvious inconsistency from other build[x] build environments (I've tested using my local Windows environment, remote ubuntu VM's, other CI tooling) just to be sure.

#### Steps to reproduce this issue

I used a whitelist-style [`.dockerignore`](https://github.com/syntaqx/api/blob/fd5e4b3c616ec3033a391734993abb92592d2403/.dockerignore) to remove any rule parsing inconsistencies that I could, trying to keep it simple:

```
*
*/

# do not ignore source files and assets - whitelist
!assets/
!cmd/
!internal/
!pkg/
!static/
!tools/

# do not ignore go module files
!go.mod
!go.sum
```

#### Expected behaviour

We should respect `.dockerignore` - He's tryin his best.

#### Actual behaviour

The [run output](https://github.com/syntaqx/api/runs/1271439413#step:10:115) shows a lot of files that aren't whitelisted, highlighted as `-` diffs. I do not see these files in other build environments using either `docker build` or `docker buildx build`

```diff
#11 [builder 6/8] COPY . .
#11 DONE 0.1s
#12 [builder 7/8] RUN ls -alh
#12 0.082 total 108K   
#12 0.083 drwxr-xr-x    1 root     root        4.0K Oct 18 15:18 .
#12 0.083 drwxrwxrwx    1 root     root        4.0K Oct 18 15:17 ..
- #12 0.083 -rw-rw-rw-    1 root     root         156 Oct 18 15:17 .dockerignore
- #12 0.083 -rw-rw-rw-    1 root     root         682 Oct 18 15:17 .editorconfig
- #12 0.083 -rw-rw-rw-    1 root     root          68 Oct 18 15:17 .envrc
- #12 0.083 -rw-rw-rw-    1 root     root        1.3K Oct 18 15:17 .gitattributes
- #12 0.083 drwxrwxrwx    3 root     root        4.0K Oct 18 15:17 .github
[...]

PR #408, opened 2021-07-13
Detail .dockerignore is ignored in git context

dvalentiate:
Add a note to README.md that .dockerignore is not processed by default by build-push-action@v2.

This issue was raised in this issue [Inconsistent dockerignore behavior from other builders](https://github.com/docker/build-push-action/issues/182) but no action was taken.

My team lost hours because `.dockerignore` is ignored unless `context: .`. This behaviour is different from expectations of all my team, and now we must update many projects. I hope this PR can help others avoid wasting so much time on an unexpected situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ By default, this action uses the [Git context](#git-context) so you don't need t
 done directly by buildkit. The git reference will be based on the [event that triggered your workflow](https://docs.github.com/en/actions/reference/events-that-trigger-workflows)
 and will result in the following context: `https://github.com/<owner>/<repo>.git#<ref>`.
 
-Be careful because **any file mutation in the steps that precede the build step will be ignored** since
+Be careful because **any file mutation in the steps that precede the build step will be ignored, including processing of the `.dockerignore` file** since
 the context is based on the git reference. However, you can use the [Path context](#path-context) using the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ By default, this action uses the [Git context](#git-context) so you don't need t
 done directly by buildkit. The git reference will be based on the [event that triggered your workflow](https://docs.github.com/en/actions/reference/events-that-trigger-workflows)
 and will result in the following context: `https://github.com/<owner>/<repo>.git#<ref>`.
 
-Be careful because **any file mutation in the steps that precede the build step will be ignored** since
+Be careful because **any file mutation in the steps that precede the build step will be ignored, including processing of the .dockerignore file** since
```
crazy-max, line 56:
```suggestion
Be careful because **any file mutation in the steps that precede the build step will be ignored, including processing of the `.dockerignore` file** since
```

crazy-max (changes requested):
~Also check DCO~

dvalentiate:
@crazy-max thanks for the review. I added your suggestion of "\`" around `.dockerignore` and rebased.

crazy-max (approved):
LGTM thanks

PR #426, opened 2021-08-05
Add note to sanitize tags

### LINKED ISSUE #424: Repository/image name should be lowercased automatically in cache parameters.
It appear the registry name is not lower-cased when used in the `cache-from` parameter. This appears related to https://github.com/docker/build-push-action/issues/37, but not exactly the same.

#### Steps to reproduce this issue

1. use ${{ github.repository}} in the `cache-from` parameter.

#### Expected behaviour

Would expect caching to work

#### Actual behaviour

> The build output this while trying to restore the cache:

```
#10 importing cache manifest from ghcr.io/My-Org/repo:main
#10 ERROR: invalid reference format: repository name must be lowercase
```

### Configuration

(This is an excerpt)

```yml
  env:
    REGISTRY: ghcr.io
    IMAGE_NAME: ${{ github.repository }}

....

      - name: Build and push dill to Docker
        with:
          context: .
          builder: ${{ steps.buildx.outputs.name }}
          file: ./docker/Dockerfile
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:main
          cache-to: type=inline,mode=max
```

crazy-max:
Closes #424

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -3,6 +3,7 @@
 * [Cannot push to a registry](#cannot-push-to-a-registry)
   * [BuildKit container logs](#buildkit-container-logs)
   * [With containerd](#with-containerd)
+* [`repository name must be lowercase`](#repository-name-must-be-lowercase)
 
 ## Cannot push to a registry
 
@@ -14,9 +15,11 @@ While pushing to a registry, you may encounter these kinds of issues:
 * `failed commit on ref "manifest-sha256:...": unexpected status: 401 Unauthorized`
 * `unexpected response: 401 Unauthorized`
 
-These issues are not directly related to this action but are rather linked to [buildx](https://github.com/docker/buildx),
-[buildkit](https://github.com/moby/buildkit), [containerd](https://github.com/containerd/containerd) or the registry
-on which you're pushing your image. The quality of error message depends on the registry and are usually not very informative.
+These issues are not directly related to this action but are rather linked to
+[buildx](https://github.com/docker/buildx), [buildkit](https://github.com/moby/buildkit),
+[containerd](https://github.com/containerd/containerd) or the registry on which
+you're pushing your image. The quality of error message depends on the registry
+and are usually not very informative.
 
 ### BuildKit container logs
 
@@ -25,8 +28,9 @@ action step and attach BuildKit container logs to your issue.
 
 ### With containerd
 
-Next you can test pushing with [containerd action](https://github.com/crazy-max/ghaction-setup-containerd) using the
-following workflow. If it works then open an issue on [buildkit](https://github.com/moby/buildkit) repository.
+Next you can test pushing with [containerd action](https://github.com/crazy-max/ghaction-setup-containerd)
+using the following workflow. If it works then open an issue on [buildkit](https://github.com/moby/buildkit)
+repository.
 
 ```yaml
 name: containerd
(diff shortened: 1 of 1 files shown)


PromoFaux:
Just a note - and maybe I was doing it wrong somehow - but this didn't actually work for me. I kept getting:

image


I ended up using:

```
- name: lowercase repo name for tag
        uses: ASzc/change-string-case-action@v1
        id: reponame
        with:
          string: ghcr.io/${{ github.repository }}
```

and referencing as : `${{ steps.reponame.outputs.lowercase }}`

crazy-max:
@PromoFaux Looking at [your workflow](https://github.com/PromoFaux/fintech-to-ynab/blob/dc98f04a21c8ae02cf3c528af7715a0759efd10e/.github/workflows/docker.yml), I suggest to use the [metadata action](https://github.com/docker/metadata-action) which will automatically sanitize tags:

```yaml
name: Docker

on:
  push:
    branches: main

jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Set up QEMU
        uses: docker/setup-qemu-action@v1

      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1

      - name: Cache Docker layers
        uses: actions/cache@v2.1.6
        with:
          path: /tmp/.docker-buildx-cache
          key: ${{ runner.os }}-docker-buildx-${{ github.sha }}
          restore-keys: ${{ runner.os }}-docker-buildx-

      - name: Login to Github Packages
        uses: docker/login-action@v1
        with:
          registry: ghcr.io
          username: ${{ github.repository_owner }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Docker meta
        id: meta
        uses: docker/metadata-action@v3
        with:
          images: ghcr.io/${{ github.repository }}
          tags: |
            sha_${{ github.sha }}
            latest

      - name: Build
        id: docker_build
        with:
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
          cache-from: type=local,src=/tmp/.docker-buildx-cache
          cache-to: type=local,dest=/tmp/.docker-buildx-cache
```

PromoFaux:
Thanks for the tips, I'll take a look!

crazy-max:
@PromoFaux 

> but this didn't actually work for me.

And yes indeed there is an issue with the syntax. It should be: `script: return '${{ github.repository }}'.toLowerCase()`. Will fix that in a follow-up and also add a note about the metadata action to sanitize tags.

PromoFaux:
AH. I came _really_ close to that solution myself! Rather I tried `${{ github.repository.toLowerCase() }}`, which also didn't work - so I gave up and found another action. I'm not lazy, honest.

PromoFaux:
Also, yeah, metadata action works well! I am already using it over on pi-hole/docker-pi-hole - I don't know why I didn't use it in this case.

crazy-max:
Yeah you can do a bunch of thing with the metadata-action. For example you could merge both nightly and release workflows and the metadata action will take care of tagging properly based on [scheduled events](https://github.com/docker/metadata-action#typeschedule), [git tag](https://github.com/docker/metadata-action#tags-input) and so on.

PR #428, opened 2021-08-05
Improve documentation around the `ssh` parameter

johnhamelink:
This change adds clarity on how to use the `ssh` parameter to pass in either an SSH agent socket, or to a file. This is obvious if you've used docker build manually, but not so much if you aren't familiar with how the command line options work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Following inputs can be used as `step.with` keys
 | `push`              | Bool     | [Push](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#push) is a shorthand for `--output=type=registry` (default `false`) |
 | `secrets`           | List     | List of secrets to expose to the build (eg. `key=string`, `GIT_AUTH_TOKEN=mytoken`) |
 | `secret-files`      | List     | List of secret files to expose to the build (eg. `key=filename`, `MY_SECRET=./secret.txt`) |
-| `ssh`               | List     | List of SSH agent socket or keys to expose to the build |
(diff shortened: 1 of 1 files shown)


crazy-max:
I think the current description is self-explanatory but we could add an example for [`--ssh` flag on buildx repo](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md) and add a link in this README when it's done.

johnhamelink:
@crazy-max See here https://github.com/docker/buildx/pull/735

flexchar:
I'm using docker build manually and pass ssh key as such `--ssh default=${SSH_AUTH_SOCK}`. In Github I have ssh private key saved as a secret. Right now it works:
```
eval "$(ssh-agent -s)"
echo "${SSH_PRIVATE_KEY}" | ssh-add -
```

But I would like to use this action and can't quite get it working. Could I have an example?

PeterLappo:
This worked for me

Setup
```yaml
      - name: Setup SSH Agent 
        env:
            SSH_AUTH_SOCK: /tmp/ssh_agent.sock
        run: |
          eval "$(ssh-agent -a $SSH_AUTH_SOCK > /dev/null)"
          ssh-add - <<< '${{ secrets.PRIVATE_SSH_KEY }}'
```
Usage
```yaml
      - name: Build and push Docker image
        uses: docker/build-push-action@v3
        env:
          SSH_AUTH_SOCK: /tmp/ssh_agent.sock
        with:
          context: .
          ssh:
            default=${{ env.SSH_AUTH_SOCK }}
```

PR #430, opened 2021-08-06
Simplified and unified argument collection

fearphage:
That's it. Just a bit of clean up.

### DIFF
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -644,7 +644,7 @@ describe('flagMap', () => {
     expect(results).toEqual([
       ['--catpants', 'a'],
       ['--catpants', 'b'],
-      ['--catpants', 'c'],
+      ['--catpants', 'c']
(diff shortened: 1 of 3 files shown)


fearphage:
@crazy-max I think I figured it out this time.

fearphage:
@crazy-max Another time around the bin.

crazy-max:
Thanks @fearphage for taking the time to put this cleanup together, and for coming back to rebase it a few times.

I'm going to close this out since the PR is now quite stale and conflicted, and the argument handling code has moved on since this was opened. The proposed refactor also changes flag ordering in a few places, so it's not something we can merge as a cleanup-only change today.

fearphage:
> The proposed refactor also changes flag ordering in a few places, so it's not something we can merge as a cleanup-only change today.

@crazy-max Is the functionality of the flags order dependent? I'm just curious.

PR #432, opened 2021-08-06
Improve dev UX

### LINKED ISSUE #431: Prettier should be a part of the dev workflow to improve developer UX
I can build and push code without it checking or ensuring that the formatting is correct. This is suboptimal. The feedback loop should be a lot shorter than waiting for GitHub Actions to run.

fearphage:
This adds two new hooks.

1. It runs `lint-staged` (which in turn runs `npm run format-check`) before `git commit`. If the check fails, the commit will be blocked.
2. It runs `npm test` before `git push`. If a test fails, the push will fail.

Fixes #431

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+lint-staged
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm test
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,3 @@
+{
+  "*.ts": "npm run format-check"
+}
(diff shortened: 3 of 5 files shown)


fearphage:
Should I add some of the commands from https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md here?

fearphage:
Can anyone tell me why `docker buildx bake pre-checkin` produced a build artifact in a branch with no code changes? That seems odd. It seems like this should have been built in the default branch and shouldn't exist here.

crazy-max:
Containerized dev environment is sandboxed, that's why we recommend to use it instead of `yarn run ...`. Also we don't use `npm` but `yarn`.

fearphage:
> that's why we recommend to use it instead of `yarn run ...`

I'll ask again: 

> Should I add some of the commands from https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md here?

Currently both of the other commands are very fast (under 2 seconds). These build commands are a bit slower and would make the experience a bit more clunky, but I thought I would ask just the same. A failing build might be signal enough though as it was for me.

fearphage:
@crazy-max Thoughts? Concerns? Feedback of any kind?

Issue #55, opened 2020-06-06
What does `always_pull` do?


uduse:
My guess is that pulling a previously built image will facilitate layer-reusing, like using the old image as a cache. Am I right?

paullaffitte:
It looks like it what it is actually doing yes. I use this option alongside with `cache_froms: organization/image`.

missinglink:
Found my way here after reading the docs and wondering the same thing...

I read it as meaning:
> Attempt to first pull the image and use the build-cache where possible to avoid building the exact same thing twice (or parts of it)

.. although if that is correct I don't see why it would be `false` by default 🤷‍♂️

missinglink:
hmm unfortunately not, I enabled `always_pull: true` after having successfully published the same action previously and it still ran the whole build (~15mins) again.

my understanding is that the build cache is not stored in the published images anyway, so I'm not sure if that's even possible?

really not sure what the purpose of `always_pull` is 😕

missinglink:
Down the rabbit hole.. the Go code appends `--pull` to [docker build](https://docs.docker.com/engine/reference/commandline/build/), the documentation for it also just says:
> Always attempt to pull a newer version of the image

Got an answer from [this Stackoverflow post](https://stackoverflow.com/questions/58488535/whats-the-purpose-of-docker-build-pull)

Essentially `always_pull` is used for the case where you have a locally built version of that specific tag (which could be old, like an old `ubuntu:latest` on your HDD) but you'd prefer that the build step pulled the newest one before running `docker build`.

So for GH actions it's not relevant, as either you've published the baseimage in a previous step (meaning it would be pulled down again) or you don't have the baseimage locally and it would be pulled as part of `docker build`.

IMO it could probably be removed, or at least better documented since we both assumed the wrong thing.

crazy-max:
@missinglink 

> IMO it could probably be removed, or at least better documented since we both assumed the wrong thing.

This input has been renamed `pull` in [build-push-action v2](https://github.com/docker/build-push-action/tree/v2-build-push) (#92) to be more consistent. You can already try it by combining it with:

* https://github.com/docker/setup-buildx-action
* https://github.com/docker/setup-qemu-action
* https://github.com/docker/login-action

crazy-max:
Version 2 has been [merged to the main branch](https://github.com/docker/build-push-action#readme) and is therefore available via `uses: docker/build-push-action@v2` (mutable tag).

[...]

PR #435, opened 2021-08-16
clarify pull option

oceanBT:


crazy-max:
Don't think that's necessary. Also it does not only concern `FROM` instructions but also:
* `# syntax=<image>`
* `COPY --from=<image>`
* `RUN --mount=type=bind,from=<image>`

oceanBT:
It took me 30 minutes to figure out exactly what is intended by "the image". Others seem to have had a similar experience (#55). 

> Also it does not only concern `FROM` instructions but also:
> 
> * `# syntax=<image>`
> * `COPY --from=<image>`
> * `RUN --mount=type=bind,from=<image>`

Thanks for this advice! I have not known that yet.

How about: 

> pull all images that are referenced in the Dockerfile (e. g. `FROM`, `COPY`) during build

crazy-max:
@oceanBT 

`Always attempt to pull a newer version of all images that are referenced in the Dockerfile` sgtm.

Might need [changes upstream](https://github.com/docker/buildx/blob/c643c2ca954c19f2a73601745e14d69967bef583/commands/build.go#L367) to keep it consistent. PTAL @tonistiigi @thaJeztah

thaJeztah:
LOL, I started to write a comment, but opened a can of worms while writing it. Anyway; here's what I wrote 😅 


> Always attempt to pull a newer version of all images that are referenced in the Dockerfile

SGTM; perhaps "newer versions"

```
Always attempt to pull newer versions of images that are referenced in the Dockerfile.
```

Unless we want to make it more generic (not mention Dockerfile?)

```
Always attempt to pull newer versions of the images images used by the build.
```


Also looking a bit at "attempt to" (it's a bit ambiguous, as it may mean either "try to pull, but don't error if it fails", or "if a newer image is available, pull it")

```
Always pull newer versions of the images images used by the build.
```

Which brings it to "what is _newer_ ? " (because it will pull the same tag, not "update to a newer version") 🤔

README.md
```diff
@@ -210,7 +210,7 @@ Following inputs can be used as `step.with` keys
 | `target`            | String   | Sets the target stage to build |
 
 ### outputs
-
+
```
crazy-max, line 213: Revert this change

crazy-max (changes requested):
Check DCO (sign-off is missing)

README.md
```diff
@@ -201,7 +201,7 @@ Following inputs can be used as `step.with` keys
 | `no-cache`          | Bool     | Do not use cache when building the image (default `false`) |
 | `outputs`           | List     | List of [output destinations](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#output) (format: `type=local,dest=path`) |
 | `platforms`         | List/CSV | List of [target platforms](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#platform) for build |
-| `pull`              | Bool     | Always attempt to pull a newer version of the image (default `false`) |
```
crazy-max, line 204: Please add back

```
(default `false`)
```

crazy-max (changes requested):
And squash your commits, thanks.

PR #443, opened 2021-08-24
fix typo

alextes:


crazy-max:
@alextes Missing DCO

alextes:
@crazy-max nothing against you personally max but I find this quite silly. For a moment I considered whether going through the hassle of cloning (I used the web UI for this), rewriting a commit, force pushing, and pinging you here was worth the effort for what is a simple typo fix.

After reading about DCO for a couple of minutes I'm also far from convinced it contributes anything. Even if I sign off, that doesn't mean the code I contribute is _actually_ code you can use and I'm no lawyer but I'd be surprised if "but he said it was code we could use so we're absolved of any responsibility" is a valid legal defense. This sounds a lot like legal trying to be useful but instead of asking "how can we enable you", be busy with "how can we look useful protecting us". Did docker ever get sued for code they used and then pay damages that would've been prevented with a DCO?? Hiiiiighly doubt it.

Anyway, rant over, doubt you have anything to do with it @crazy-max 😂 , have a nice day, but who knows someday someone up the chain might ask you about how to get more contributions to come in and you can forward this :p

PR #453, opened 2021-08-30
Composite action

### LINKED ISSUE #208: Feedback for version 2

crazy-max:
Fixes #208

Try-out the composite feature which now supports `uses:` keyword:
* Set up Buildx and QEMU
* Cache
* Metadata
* Build and push

Usage:

```yaml
name: ci

on:
  push:
    branches:
      - 'master'

jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v2
      -
        name: Login to DockerHub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      -
        name: Build with composite
        uses: docker/build-push-action/composite@v2
        with:
          context: .
          images: |
            user/app
          platforms: linux/amd64,linux/arm64
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -673,3 +673,29 @@ jobs:
         name: Dump context
         if: always()
         uses: crazy-max/ghaction-dump-context@v1
+
+  composite:
+    runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Build with composite
+        uses: ./composite
+        with:
+          context: .
+          file: ./test/Dockerfile
+          images: |
+            localhost:5000/name/app
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+      -
+        name: Dump context
+        if: always()
+        uses: crazy-max/ghaction-dump-context@v1
--- a/composite/action.yml
+++ b/composite/action.yml
@@ -1,0 +1,69 @@
+name: 'Docker composite action'
+description: 'Set up Buildx and QEMU + Cache + Metadata + Build and push'
+
+inputs:
+  images:
+    description: "List of Docker image slugs (eg. name/app)"
+    required: true
+  context:
+    description: "Build's context is the set of files located in the specified PATH or URL. (default Git context)"
+    required: false
+  file:
+    description: "Path to the Dockerfile. (default {context}/Dockerfile)"
+    required: false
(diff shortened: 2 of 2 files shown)


crazy-max:
We would prefer reusable workflows for such case, see #1321

PR #486, opened 2021-10-15
Added 'quiet' docker arg support

jj-mino:


README.md
```diff
@@ -210,6 +210,7 @@ Following inputs can be used as `step.with` keys
 | `ssh`               | List     | List of SSH agent socket or keys to expose to the build |
 | `tags`              | List/CSV | List of tags |
 | `target`            | String   | Sets the target stage to build |
+| `quiet`             | String   | Suppress the build output and print image ID on success |
```
fearphage, line 213: Looks like the previous options were in alphabetical order.

src/context.ts
```diff
@@ -26,6 +26,7 @@ export interface Inputs {
   noCache: boolean;
   outputs: string[];
   platforms: string[];
+  quiet: boolean;
```
fearphage, line 29: Alphabetize here too.

src/context.ts
```diff
@@ -74,6 +75,7 @@ export async function getInputs(defaultContext: string): Promise<Inputs> {
     load: core.getBooleanInput('load'),
     network: core.getInput('network'),
     noCache: core.getBooleanInput('no-cache'),
+    quiet: core.getBooleanInput('quiet'),
```
fearphage, line 78: Alphabetize here too.

fearphage (changes requested):
You should try to maintain the current state of things.

crazy-max:
As discussed in #501, we are not going to add the `quiet` flag as it does not make much sense for the action itself.

PR #500, opened 2021-11-16
readme: remove v1 section

crazy-max:
More than a year has passed since v2 so time to remove v1 section from README. Not sure about removing `UPGRADE.md` though. cc @tonistiigi @thaJeztah @chris-crone 


chris-crone (approved):
SGTM

PR #501, opened 2021-11-16
add `cgroup-parent`, `shm-size`, `ulimit` inputs

crazy-max:
available since [buildx 0.7.0](https://github.com/docker/buildx/releases/tag/v0.7.0).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,6 +336,86 @@ jobs:
         if: always()
         uses: crazy-max/ghaction-dump-context@v1
 
+  shm-size:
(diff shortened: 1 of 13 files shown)


tonistiigi (commented):
I don't think it makes sense to add quiet this way. It's only for backward compact anyway.

Note that these also depend on buildkit version.

crazy-max:
> I don't think it makes sense to add quiet this way. It's only for backward compact anyway.

Yes I hesitated to add it because it also brings nothing to the action as such. I will remove it.

tonistiigi (approved):

PR #522, opened 2021-12-20
Change the default branch in code examples from "master" to "main"

rezanmz:
GitHub has changed the default branch name from "master" to "main". For people who copy and paste these code snippets into their projects, it becomes problematic and confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ name: ci
 on:
   push:
     branches:
-      - 'master'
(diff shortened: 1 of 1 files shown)


crazy-max (changes requested):
We can simply remove them so we have:

```yaml
on:
  push:

jobs:
  ...
```

I think it's ok as an example.

PR #527, opened 2021-12-23
Update README.md

rfrm:
It's not clear that the difference when using the path context compared to using the git context is that you need to use the checkout action.


crazy-max:
@rfrm I don't think we need it as this is already specified in the [Usage section](https://github.com/docker/build-push-action#usage).

PR #531, opened 2021-12-28
Add subdirectory for Git context

### LINKED ISSUE #460: Feature request: Allow specifying subdirectory with default git context
When using git repository contexts, Docker allows specifying a subdirectory using the syntax `<repo>#<ref>:<subdir>`, as documented here:

* https://docs.docker.com/engine/reference/commandline/build/#git-repositories

BuildKit used to not support this feature, but as of BuildKit 0.9.0, it does:

* Release notes: https://github.com/moby/buildkit/releases/tag/v0.9.0
* https://github.com/moby/buildkit/issues/1684
* https://github.com/moby/buildkit/pull/2116

Can this action add support for this, to avoid the need to use checkouts to build subdirectories?

### Workaround

For the time being, specifying a full git context seems to work; for example, to build subdirectory `spam`:

```yaml
        with:
          context: "https://github.com/${{ github.repository }}.git#${{ github.sha }}:spam"
```

However, this doesn't use the same logic for determining the default git context as the action does, so it might be fragile.

### LINKED ISSUE #528: Action doesn't perform git checkout
#### Expected behaviour

According to README:

> By default, this action uses the Git context so you don't need to use the actions/checkout action to checkout the repository because this will be done directly by buildkit. 

#### Actual behaviour

Checkout doesn't happen.

### Configuration

This

```yml
      - name: Build and push container
        with:
          context: ./app
          push: true
          tags: ${{ env.ECR_REGISTRY }}/proxyco/madmax-api:stable
```

fails with error:

```
/usr/local/bin/docker buildx build --tag ***.dkr.ecr.us-west-2.amazonaws.com/proxyco/madmax-api:stable --iidfile /tmp/docker-build-push-s6cfCu/iidfile --metadata-file /tmp/docker-build-push-s6cfCu/metadata-file --push ./app
error: unable to prepare context: path "./app" not found
Error: buildx failed with: error: unable to prepare context: path "./app" not found
```

I have to explicitly do a checkout first, for the build to be successful.

```yml
      - uses: actions/checkout@v2

      - name: Build and push container
        with:
          context: ./app
          push: true
          tags: ${{ env.ECR_REGISTRY }}/proxyco/madmax-api:stable
```

It's a private repo.

This is run on a self-hosted runner using ubuntu-latest.

What am I doing wrong?

BeyondEvil:
Since v0.9.0 of BuildKit (BuildX v0.7.0) you can provide a subdirectory
to the default Git context.

Closes #460
Closes #528

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,19 @@ By default, this action uses the [Git context](#git-context) so you don't need t
 done directly by buildkit. The git reference will be based on the [event that triggered your workflow](https://docs.github.com/en/actions/reference/events-that-trigger-workflows)
 and will result in the following context: `https://github.com/<owner>/<repo>.git#<ref>`.
 
+You can provide a subdirectory to the [Git context](#git-context) by using the following [Handlebars template](https://handlebarsjs.com/guide/) expression `{{defaultContext}}`:
+
+```yaml
+      -
+        name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: {{defaultContext}}:docker
+          push: true
+          tags: user/app:latest
+```
+
 Be careful because **any file mutation in the steps that precede the build step will be ignored, including processing of the `.dockerignore` file** since
 the context is based on the git reference. However, you can use the [Path context](#path-context) using the
 [`context` input](#inputs) alongside the [`actions/checkout`](https://github.com/actions/checkout/) action to remove
--- a/src/context.ts
+++ b/src/context.ts
@@ -9,6 +9,7 @@ import {issueCommand} from '@actions/core/lib/command';
 import * as github from '@actions/github';
 
 import * as buildx from './buildx';
+import * as handlebars from 'handlebars';
 
 let _defaultContext, _tmpDir: string;
 
@@ -97,7 +98,7 @@ export async function getArgs(inputs: Inputs, defaultContext: string, buildxVers
   let args: Array<string> = ['buildx'];
   args.push.apply(args, await getBuildArgs(inputs, defaultContext, buildxVersion));
   args.push.apply(args, await getCommonArgs(inputs, buildxVersion));
-  args.push(inputs.context);
+  args.push(handlebars.compile(inputs.context)({defaultContext}));
   return args;
 }
 
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -491,6 +491,24 @@ nproc=3`],
         '.'
       ]
     ],
+    [
+      15,
+      '0.7.0',
+      new Map<string, string>([
+        ['context', '{{defaultContext}}:docker'],
+        ['load', 'false'],
+        ['no-cache', 'false'],
+        ['push', 'false'],
+        ['pull', 'false'],
+      ]),
+      [
+        'buildx',
+        'build',
+        '--iidfile', '/tmp/.docker-build-push-jest/iidfile',
+        '--metadata-file', '/tmp/.docker-build-push-jest/metadata-file',
+        'https://github.com/docker/build-push-action.git#refs/heads/test-jest:docker'
+      ]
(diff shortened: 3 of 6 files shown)


src/context.ts
```diff
@@ -97,7 +97,11 @@ export async function getArgs(inputs: Inputs, defaultContext: string, buildxVers
   let args: Array<string> = ['buildx'];
   args.push.apply(args, await getBuildArgs(inputs, defaultContext, buildxVersion));
   args.push.apply(args, await getCommonArgs(inputs, buildxVersion));
+  if (inputs.context.startsWith('{{defaultContext}}') && buildx.satisfies(buildxVersion, '>=0.7.0')) {
```
BeyondEvil, line 100: Should I add any type of error-handling here?

As it stands, if the version is <0.7.0 it's going to set the context to (literal) `{{defaultContext}}:docker`.

BeyondEvil:
Locally, `test` and `pre-checkin` passes, but not `validate`.

```sh
 => ERROR [build-validate build-validate 1/1] RUN --mount=type=bind,target=.,rw <<EOT (set -e...)                                                                                                                                                                                                                                           0.5s
------
 > [build-validate build-validate 1/1] RUN --mount=type=bind,target=.,rw <<EOT (set -e...):
#25 0.383 ERROR: Build result differs. Please build first with "docker buildx bake build"
#25 0.390 M  dist/index.js
------
error: failed to solve: executor failed running [/bin/sh -c set -e
git add -A
cp -rf /out/* .
if [ -n "$(git status --porcelain -- dist)" ]; then
  echo >&2 'ERROR: Build result differs. Please build first with "docker buildx bake build"'
  git status --porcelain -- dist
  exit 1
fi
]: exit code: 1
```

Not sure how to fix it.

src/context.ts
crazy-max: I don't think we have to check buildx version here.

src/context.ts
```diff
@@ -97,7 +97,11 @@ export async function getArgs(inputs: Inputs, defaultContext: string, buildxVers
...
+    inputs.context = inputs.context.replace('{{defaultContext}}', defaultContext);
```
crazy-max, line 101: I think it would be better to use [handlebars template](https://handlebarsjs.com/guide/) like it was done in our [metadata action](https://github.com/docker/metadata-action#global-expressions).

src/context.ts
BeyondEvil: Because the version of build-push-action is tied to the BuildX version?

crazy-max: No because it only affects buildkit not buildx. Also we need to vendor buildkit on moby https://github.com/moby/moby/pull/42968 otherwise it will not work with the docker driver.

src/context.ts
BeyondEvil: Updated @crazy-max 👍

README.md
```diff
@@ -43,6 +43,19 @@ By default, this action uses the [Git context](#git-context) so you don't need t
 done directly by buildkit. The git reference will be based on the [event that triggered your workflow](https://docs.github.com/en/actions/reference/events-that-trigger-workflows)
 and will result in the following context: `https://github.com/<owner>/<repo>.git#<ref>`.
 
+Beginning with BuildKit version `v0.9.0` ([Buildx](https://github.com/docker/buildx) `v0.7.0`) you can provide a subdirectory to the [Git context](#git-context) by using the magic variable `{{defaultContext}}`:
```
BeyondEvil, line 46: Should I replace `magic variable` with `handlebar template`?

crazy-max: Yes you can write smth like https://github.com/docker/metadata-action#global-expressions

BeyondEvil: Updated @crazy-max 👍

crazy-max (approved):
LGTM

PR #532, opened 2021-12-28
Add context-subdir input

### LINKED ISSUE #460: Feature request: Allow specifying subdirectory with default git context

### LINKED ISSUE #528: Action doesn't perform git checkout

actualben:
This is an alternative to #531, that PR implements the magic var which you described in https://github.com/docker/build-push-action/issues/460#issuecomment-917000054 .

This approach instead uses a new input `context-subdir` which gets appended to the context URL. This new input is ignored (with a warning) if the user also specifies a `context` input such (as a path). The potential benefit to this approach is that it might be a little easier for users than utilizing a magic var.

magic var:
```yaml
name: Build and push
id: docker_build
with:
  push: true
  tags: user/app:latest
  context: {{defaultContext}}:docker
```

vs

new input:
```yaml
name: Build and push
id: docker_build
with:
  push: true
  tags: user/app:latest
  context-subdir: docker
```

Closes #460
Closes #528

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -20,6 +20,7 @@ export interface Inputs {
   cacheTo: string[];
   cgroupParent: string;
   context: string;
+  contextSubdir: string;
   file: string;
   labels: string[];
   load: boolean;
@@ -73,6 +74,7 @@ export async function getInputs(defaultContext: string): Promise<Inputs> {
     cacheTo: await getInputList('cache-to', true),
     cgroupParent: core.getInput('cgroup-parent'),
     context: core.getInput('context') || defaultContext,
+    contextSubdir: core.getInput('context-subdir'),
     file: core.getInput('file'),
     labels: await getInputList('labels', true),
     load: core.getBooleanInput('load'),
@@ -97,7 +99,15 @@ export async function getArgs(inputs: Inputs, defaultContext: string, buildxVers
   let args: Array<string> = ['buildx'];
   args.push.apply(args, await getBuildArgs(inputs, defaultContext, buildxVersion));
   args.push.apply(args, await getCommonArgs(inputs, buildxVersion));
-  args.push(inputs.context);
+  let context: string = inputs.context;
+  if (inputs.contextSubdir) {
+    if (context == defaultContext) {
+      context = `${context}:${inputs.contextSubdir}`;
+    } else {
+      core.warning('"context-subdir" input is ignored when a "context" input is present');
+    }
+  }
+  args.push(context);
   return args;
 }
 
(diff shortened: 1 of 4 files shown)


src/context.ts
```diff
@@ -97,7 +99,15 @@ export async function getArgs(inputs: Inputs, defaultContext: string, buildxVers
   let args: Array<string> = ['buildx'];
   args.push.apply(args, await getBuildArgs(inputs, defaultContext, buildxVersion));
   args.push.apply(args, await getCommonArgs(inputs, buildxVersion));
-  args.push(inputs.context);
+  let context: string = inputs.context;
+  if (inputs.contextSubdir) {
```
BeyondEvil, line 103: Since this is only available in BuildX starting with `v0.7.0`, shouldn't you check that `buildxVersion` satisfies that? 🤔

actualben:
closing in favor of #531

PR #560, opened 2022-02-13
Add an example of accessing the secrets file without root permissions

efernandesng:


crazy-max:
@efernandesng I think this example should be in the buildx repo instead: https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#secret

crazy-max:
cc @dvdksn

dvdksn:
Sweet update. I agree. @efernandesng  Could you raise a PR that adds a section to the doc linked by @crazy-max above?

PR #613, opened 2022-05-09
docs: clarify push vs. load limitation

BretFisher:
Adding a line to each option `load` and `push` to clarify that they can't be used together.


crazy-max (changes requested):
Can you rebase and squash please? Otherwise LGTM!

PR #649, opened 2022-07-09
Update README.md

AlphaWong:
update doc

Signed-off-by: Alpha <AlphaWong@users.noreply.github.com>


jedevc:
Do you have any more context for this?

Dockerfile filenames and extensions can be anything you like - while it's common practice to end in `.Dockerfile`, it's definitely not a requirement.

AlphaWong:
> Do you have any more context for this?
> 
> Dockerfile filenames and extensions can be anything you like - while it's common practice to end in `.Dockerfile`, it's definitely not a requirement.

As I refer to the Docker doc
```console
 docker build -f dockerfiles/Dockerfile.debug -t myapp_debug .
 docker build -f dockerfiles/Dockerfile.prod  -t myapp_prod .
```
https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f

# before I change the file name
## Github action
image
## result
image

# after I change the file name
## Github action
image
## result
image

# what I think
`docker buildx build` does not share the same behavior like docker build

@jedevc

AlphaWong:
My bad. fucking macos is case insensitive as I put `Dockerfile.dev` it does not consider it us a new file.
And Ubuntu cannot find `Dockerfile.dev` as the repo only has `dockerfile.dev`

Sorry to boring u @jedevc 

It is my bad

PR #653, opened 2022-07-18
`no-cache-filters` input

### LINKED ISSUE #645: Feature request: Add no-cache-filter
This is especially for nightly builds, I'd like to be able to pass the `--no-cache-filter` flag to force the later stages to re-run (but still cache the earlier ones)

crazy-max:
fixes #645 


crazy-max:
> See comment [here](https://github.com/docker/buildx/pull/1215#issuecomment-1187626983)

Yes indeed, not sure myself as we currently make inputs plural in this action like `outputs`.

jedevc:
> we currently make inputs plural in this action like `outputs`

Ah sorry, I got confused. Makes sense to allow this field as this then, but long-term may be worth discussing about option consistency throughout actions/buildx/bake.

jedevc (approved):

PR #657, opened 2022-07-27
Replace deprecated `fs.rmdir` with `fs.rm`

bendrucker:
Fixes the following deprecation warning during cleanup:

```
(node:7106) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

https://nodejs.org/api/deprecations.html#DEP0147


crazy-max (approved):
LGTM thanks

PR #661, opened 2022-08-03
Update comment regarding the support of subdir context

dud225:


jedevc:
Not sure if I understand the reason for this 😕

The README already describes that you should use the setup-buildx-action, not sure if we need to call it out again below!

dud225:
I'm personally used to omit this addittional action because for the basic usage the Buildkit embedded within the Docker engine just works fine. I've removed the mention to it.

crazy-max:
Thank you for contributing! It appears your commit message is missing a DCO sign-off, causing the DCO check to fail.

We require all commit messages to have a Signed-off-by line with your name and e-mail, which looks something like:

```
Signed-off-by: YourFirsName YourLastName <yourname@example.org>
```

There is no need to open a new pull request, but to fix this (and make CI pass), you need to amend the commit(s) in this pull request, and "force push" the amended commit.

You can find some instructions in the output of the DCO check (which can be found in the "checks" tab on this pull request), as well as in the [Moby contributing guide](https://github.com/moby/moby/blob/v20.10.11/docs/contributing/set-up-git.md).

Steps to do so "roughly" come down to:

1. Set your name and e-mail in git's configuration:
   ```
   git config --global user.name "YourFirstName YourLastName"
   git config --global user.email "yourname@example.org"
   ```
   (Make sure to use your real name (not your GitHub username/handle) and e-mail)

2. Clone your fork locally

3. Check out the branch associated with this pull request

4. Sign-off and amend the existing commit(s)
   ```
   git commit --amend --no-edit --signoff
   ```
   If your pull request contains multiple commits, either squash the commits (if needed) or sign-off each individual commit.

5. _Force push_ your branch to GitHub (using the `--force` or [`--force-with-lease`](https://stackoverflow.com/questions/52823692/git-push-force-with-lease-vs-force) flags) to update the pull request.

Sorry for the hassle, and let me know if you need help or more detailed instructions!

crazy-max (approved):
LGTM thanks

PR #665, opened 2022-08-05
Docs: comment on availability of built images in other build steps

baibaratsky:


goitf (approved):
thank you

crazy-max:
This is already the case when using `load: true`.

baibaratsky:
@crazy-max Just `load: true` doesn't allow you to use the built image in the next build step with the container driver. This is the point.

docs/advanced/export-docker.md
```diff
@@ -33,3 +33,45 @@ jobs:
         run: |
           docker image inspect myimage:latest
```
crazy-max, line 34: > Just `load: true` doesn't allow you to use the built image in the next build step with the container driver. This is the point.

If this step works then it's loaded to docker store.

baibaratsky: Yes, it works, and it's loaded. But it isn't available in the next step in the building container when using default driver.

crazy-max: Oh you mean reusing this image as base with another build-push-action step. Then yes it will not be visible by the container builder. Was not clear to me you were using another context and Dockerfile.  Best way for this case is to use `build-contexts` input like:

```yaml
      -
        name: Build base image
        uses: docker/build-push-action@v3
        with:
          context: base
          load: true
          tags: my-base-image:latest
      -
        name: Build image from my-base-image:latest
        uses: docker/build-push-action@v3
        with:
          context: .
          build-contexts: |
            alpine=docker-image://my-base-image:latest
          tags: myimage:latest
```

Where `alpine` is the name of the base image (or stage name if specified) in your dockerfile like:

```dockerfile
FROM alpine
```

This way you can keep the container builder that provides all the BuildKit features but the `docker` driver doesn't: https://docs.docker.com/build/buildx/drivers/

More info: https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context

I think you could also use bake to ease that so you would need only one step that will build first the base image and use its result for the main one: https://docs.docker.com/build/bake/build-contexts/#using-a-result-of-one-target-as-a-base-image-in-another-target

```hcl
# docker-bake.hcl
target "base" {
  dockerfile = "baseapp.Dockerfile"
}

target "app" {
  contexts = {
    baseapp = "target:base"
  }
}
```

```yaml
      -
        name: Build
        uses: docker/bake-action@v2
        with:
          target: app
```

baibaratsky: I use the base image to build two different images from it, so the first context option looks more suitable.

baibaratsky:
@crazy-max I still think it would be useful to have these comments in the “export to Docker” section. It could save a lot of time to others.

crazy-max:
@baibaratsky Thanks for the changes. I think in a follow-up we might create a new page to reuse build result as it's a bit out of scope with "Export image to Docker" page. But in the meantime that sounds good to me.

crazy-max:
Pushed extra commit with a dedicated page about named contexts. Removed the bake example that will be added in [`bake-action` repo](https://github.com/docker/bake-action) instead.

crazy-max (approved):

PR #666, opened 2022-08-06
use ${{ env.TEST_TAG }} to tag image in push step

deronnax:
Signed-off-by: Mathieu Dupuy <deronnax@gmail.com>


crazy-max (approved):
LGTM thanks

PR #672, opened 2022-08-17
ci: fix workflow dispatch inputs

crazy-max:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,10 @@ on:
     branches:
       - 'master'
 
+env:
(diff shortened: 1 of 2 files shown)


jedevc (approved):
This looks right to me :)

I'm confused, did this just not work before at all? It looks like `matrix` doesn't have a `buildx-version` property, so did this just get silently omitted?

crazy-max:
> I'm confused, did this just not work before at all? It looks like `matrix` doesn't have a `buildx-version` property, so did this just get silently omitted?

Yes silently omitted even if there is no matrix available :disappointed:

PR #704, opened 2022-10-12
Remove workaround for setOutput

crazy-max:


csaba-kovacs:
this fixed the save state deprecation?

PR #709, opened 2022-10-25
Document the use of scope with GHA caching

### LINKED ISSUE #706: Document scope attribute
Enhancement [per prior discussion](https://github.com/docker/build-push-action/issues/624#issuecomment-1141336628)

Still not documented in https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#github-cache

yanirs:
Resolves #706.

### DIFF
--- a/docs/advanced/cache.md
+++ b/docs/advanced/cache.md
@@ -109,6 +109,10 @@ exclusively used in a GitHub Action workflow as the `url` (`$ACTIONS_CACHE_URL`)
 and `token` (`$ACTIONS_RUNTIME_TOKEN`) attributes are populated when a workflow
 is started.
 
+When building multiple images, add `scope=<scope_name>` to `cache-from` and `cache-to` to avoid
+overwriting the cache index. See the docs for
+[GitHub Actions cache](https://docs.docker.com/build/building/cache/backends/gha/#scope).
+
 ```yaml
 name: ci
 


docs/advanced/cache.md
```diff
@@ -109,6 +109,9 @@ exclusively used in a GitHub Action workflow as the `url` (`$ACTIONS_CACHE_URL`)
 and `token` (`$ACTIONS_RUNTIME_TOKEN`) attributes are populated when a workflow
 is started.
 
+When building multiple images, add `scope=<scope_name>` to `cache-from` and `cache-to` to avoid
+overwriting the cache index. See https://github.com/moby/buildkit/issues/2885.
```
crazy-max, line 113: Better to link to https://docs.docker.com/build/building/cache/backends/gha/#scope

crazy-max, line 113: Better to link to https://docs.docker.com/build/building/cache/backends/gha/#scope

crazy-max, line 113: Better to link to https://docs.docker.com/build/building/cache/backends/gha/#scope

crazy-max (changes requested):

docs/advanced/cache.md
yanirs: Good point. Updated.

crazy-max (changes requested):
Can you squash your commits please?

yanirs:
Yeah, I actually do mind. This is a trivial change that'd help people who hit the same issue, but there was no need for it to have multiple review rounds or for it to be sitting as an open PR for a month. Specifically, you could have squashed the commits on your side or merged it after I squashed them.

I'm done here. If you want to move the extra paragraph to the other repo (which will still help people), you can do it.

PR #713, opened 2022-10-27
fix: incorrect yaml in documentation

sammcj:
Fix invalid YAML in documentation.

Previous the name key for each section had it's key identifier incorrectly on the previous line.

This may seem trivial, but if users had copy/pasted the examples given their YAML would be incorrectly formatted.


crazy-max:
Afaik this is correctly formatted. The extra new line for each step is set on purpose for better visibility. You can check on http://www.yamllint.com/ or https://codebeautify.org/yaml-validator.

Issue #557, opened 2022-02-02
Build args 


shyam-ks:
I am trying to setup docker build in Github actions

```
- name: build-1
        run: |
          docker build . --build-arg JFROG_USERNAME=${{ secrets.JFROG_CREDS_USR }} --build-arg JFROG_PWD=${{ secrets.JFROG_CREDS_PSW }} 

      - name: build-2
        uses: docker/build-push-action@v2
        with:
          context: .
          file: ./Dockerfile
          push: false
          build-args: JFROG_USERNAME=${{ secrets.JFROG_CREDS_USR }},JFROG_PWD=${{ secrets.JFROG_CREDS_PSW }}
```
I have two args inside  Dockerfile

ARG JFROG_PWD
ARG JFROG_USERNAME

### Behaviour
In Dockerfile  the command is =  value="${JFROG_USERNAME}" 

The first one (build-1) builds successfully, the Second one fails because the whole build args value is going to a single variable. like below
value="***,JFROG_PWD=***"

Expected behavior - value="***"

I'm not able to figure out why the 2 build args are not setting values to their respective args inside Dockerfile when I am using docker/build-push-action@v2

makoncline:
I think you should write it like this.
```
build-args: |
    JFROG_USERNAME=${{ secrets.JFROG_CREDS_USR }}
    JFROG_PWD=${{ secrets.JFROG_CREDS_PSW }}
```

ckingbailey:
Would be nice if this was documented. There are no examples at [the examples link](https://github.com/docker/build-push-action#examples), and [the documentation](https://github.com/docker/build-push-action#inputs) says it's a "List", which I interpreted to mean a YAML list, like
```yaml
build-args:
  - foo=bar
  - baz=yaz
```
or
```yaml
build-args: [ 'foo=bar', 'baz=yaz' ]
```
which GitHub Actions language server in VSCode tells me is not valid input. (I'm guessing only strings, numbers, and booleans are valid arguments to an action? You learn something new...)

For reference, an example of an action that does document list-like syntax like this is [actions/upload-artifact](https://github.com/actions/upload-artifact).

aharpervc:
> Would be nice if this was documented

I agree, I just found this discussion by searching for build args, since the readme offered no clarity and I assumed someone else had already complained 😁

ibrahima:
[...]

Issue #380, opened 2021-06-06
Multiple build-args not resulting in correct docker build syntax


NonsenseSynapse:
### Troubleshooting

Before sumbitting a bug report please read the [Troubleshooting doc](https://github.com/docker/build-push-action/blob/master/TROUBLESHOOTING.md).

#### Steps to reproduce this issue
Using the following Dockerfile and Github Action:
<details><summary>Dockerfile</summary>
<p>

```Dockerfile
FROM alpine:3.7

ARG FIRST_ARG
RUN echo first arg: ${FIRST_ARG}

ARG SECOND_ARG
RUN echo second arg: ${SECOND_ARG}
```

</p>
</details>

<details><summary>Github Action</summary>
<p>

```yaml
name: Build Test

on:
  workflow_dispatch:
jobs:
  build_and_publish:
    runs-on: ubuntu-latest
    steps:
      -
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
      -
        name: Login to DockerHub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      -
        name: Clone Github Repo
        uses: actions/checkout@master
      -
        name: Build and push
        id: docker_build
        uses: docker/build-push-action@v2
        with:
          context: .
          file: Dockerfile
          push: true
          tags: nonsensesynapse/app:tag
          build-args:
            FIRST_ARG="First"
            SECOND_ARG="Second"
```

</p>
</details>

#### Expected behaviour
The correct docker syntax is to have a separate `--build-arg` for each argument.
```
docker build --progress=plain --build-arg FIRST_ARG="First" --build-arg SECOND_ARG="Second" - < Dockerfile
```
which gives me the following output:
```
#5 [2/3] RUN echo first arg: First
#5 sha256:398661483d922dc4c6b2cb85f72f84de0a8765d511cd2701aa496738019b5baf
#5 0.267 first arg: First
#5 DONE 0.3s

#6 [3/3] RUN echo second arg: Second
#6 sha256:379a45924550875394a9c34cf314ddc4ca9389b749d8205e6e6d36997773d21d
#6 0.331 second arg: Second
#6 DONE 0.4s
```

Source from [docker documentation](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

[...]

PR #734, opened 2022-11-30
Update README.md

josh-gree:
I spent quite a while today trying to work out why my action was failing - this makes things a little clearer - might be worthwhile actually having one of the example contain the specification of multiple build args!

Make the `build-args` description clearer - should be specified as;
```
          build-args: |
            KEY1=FOO
            KEY2=BAR
```

and not like;

```
          build-args: KEY1=FOO,KEY2=BAR
```

Signed-off-by: Joshua Greenhalgh <joshuadouglasgreenhalgh@gmail.com>


crazy-max:
Thanks but this is already specified in the README: https://github.com/docker/build-push-action/#inputs

`build-args` is of type `List`.

> List type is a newline-delimited string
> ```yaml
> cache-from: |
>   user/app:cache
>   type=local,src=path/to/dir
> ```

See also
* https://github.com/docker/build-push-action/issues/380#issuecomment-855477968
* https://github.com/docker/build-push-action/issues/557

PR #746, opened 2022-12-19
add attests, provenance and sbom inputs

crazy-max:


src/context.ts
```diff
@@ -150,6 +161,18 @@ async function getBuildArgs(inputs: Inputs, defaultContext: string, context: str
   if (inputs.platforms.length > 0) {
     args.push('--platform', inputs.platforms.join(','));
   }
+  if (buildx.satisfies(buildxVersion, '>=0.10.0')) {
+    if (inputs.provenance) {
+      args.push('--provenance', inputs.provenance);
+    } else if (fromPayload('repository.private') !== false) {
+      args.push('--provenance', `mode=min,inline-only=true`);
+    } else {
+      args.push('--provenance', `mode=max,builder-id=${process.env.GITHUB_SERVER_URL || 'https://github.com'}/${github.context.repo.owner}/${github.context.repo.repo}/actions/runs/${github.context.runId}`);
```
crazy-max, line 170: @tonistiigi As discussed, set provenance mode max and builder id for public repos by default.

README.md
```diff
@@ -190,35 +190,38 @@ Following inputs can be used as `step.with` keys
...
+| `no-cache`         | Bool        | Do not use cache when building the image (default `false`)                                                                                                                        |
+| `no-cache-filters` | List/CSV    | Do not cache specified stages                                                                                                                                                     |
+| `outputs`¹         | List        | List of [output destinations](https://docs.docker.com/engine/reference/commandline/buildx_build/#output) (format: `type=local,dest=path`)                                         |
+| `platforms`        | List/CSV    | List of [target platforms](https://docs.docker.com/engine/reference/commandline/buildx_build/#platform) for build                                                                 |
+| `provenance`       | Bool/String | [Provenance](https://docs.docker.com/build/attestations/provenance/) is a shorthand for `--attest=type=provenance`                                                                |
```
tonistiigi, line 213: "Generate provenance attestation for the build"

README.md
```diff
@@ -190,35 +190,38 @@ Following inputs can be used as `step.with` keys
...
+| `pull`             | Bool        | Always attempt to pull all referenced images (default `false`)                                                                                                                    |
+| `push`             | Bool        | [Push](https://docs.docker.com/engine/reference/commandline/buildx_build/#push) is a shorthand for `--output=type=registry` (default `false`)                                     |
+| `sbom`             | Bool/String | [SBOM](https://docs.docker.com/build/attestations/sbom/) is a shorthand for `--attest=type=sbom`                                                                                  |
```
tonistiigi, line 216: "Generate SBOM attestation". We also need to describe the defaults. If not here then in a special block.

src/context.ts
```diff
@@ -150,6 +161,18 @@ async function getBuildArgs(inputs: Inputs, defaultContext: string, context: str
   if (inputs.platforms.length > 0) {
     args.push('--platform', inputs.platforms.join(','));
   }
+  if (buildx.satisfies(buildxVersion, '>=0.10.0')) {
```
tonistiigi, line 164: What if buildkit is older than v0.11 . We need to make sure it doesn't error by default.

src/context.ts
tonistiigi: Maybe `builder-id` can be set always?

If user sets own provenance we should still add `builder-id` unless user sets their own value.

.github/workflows/ci.yml
```diff
@@ -491,6 +491,68 @@ jobs:
...
+      -
+        name: Check output folder
+        if: matrix.target == 'binary'
+        run: |
+          tree /tmp/buildx-build
+      -
+        name: Print provenance
+        if: matrix.target == 'binary'
+        run: |
+          cat /tmp/buildx-build/provenance.json | jq
```
tonistiigi, line 549: Could we also do some basic content validation for a build parameter and sbom pkg for example.

src/context.ts
jedevc: > Maybe `builder-id` can be set always?

I think this is a good idea - maybe we could hide it if `BUILDX_GIT_INFO=false` though, so that we have a single setting for the git info?

.github/workflows/ci.yml
crazy-max: You mean schema validation?

farvour:
I am using this action in our workflows to build. I am discovering that ECR does not like the attestation layers pushed when security scanning is enabled on a repository. It results in a `Failed` on the image UI state. Not a major issue, but sounds like this new feature to this action will let me turn these off with the newer docker buildx (0.10.0+) if my understanding is correct?

README.md
lucacome, line 213: looks like the link is https://docs.docker.com/build/attestations/slsa-provenance/ ?

src/buildx.ts
```diff
@@ -126,6 +141,109 @@ export async function isAvailable(standalone?: boolean): Promise<boolean> {
...
+    })
+    .then(res => {
+      if (res.stderr.length > 0 && res.exitCode != 0) {
+        throw new Error(res.stderr.trim());
+      }
+      return parseInspect(res.stdout);
+    });
+}
+
+async function parseInspect(data: string): Promise<Builder> {
```
jedevc, line 174: This is imported from https://github.com/docker/setup-buildx-action/blob/39a1a82492fd1ad19af19d61b5f748e4cb6cd1af/src/buildx.ts#L111-L182?

LGTM for now, but is there a way of importing cross repo we could look at later?

src/context.ts
```diff
@@ -150,6 +161,22 @@ async function getBuildArgs(inputs: Inputs, defaultContext: string, context: str
...
+    if (inputs.provenance) {
+      args.push('--provenance', inputs.provenance);
+    } else if (await buildx.satisfiesBuildKitVersion(inputs.builder, '>=0.11.0', standalone)) {
+      if (fromPayload('repository.private') !== false) {
+        args.push('--provenance', `mode=min,inline-only=true`);
+      } else {
+        args.push('--provenance', `mode=max,builder-id=${process.env.GITHUB_SERVER_URL || 'https://github.com'}/${github.context.repo.owner}/${github.context.repo.repo}/actions/runs/${github.context.runId}`);
+      }
+    } else {
+      args.push('--provenance', 'false');
```
jedevc, line 174: Probably worth adding a comment explaining why we do this and what triggers this.

src/buildx.ts
crazy-max: Yes same but don't want to waste time creating a library for now. We should just support proper format for `inspect` command, see https://github.com/docker/buildx/pull/830#issuecomment-1348729303

src/context.ts
crazy-max: This is not required actually

src/context.ts
crazy-max: Now `builder-id` is always set if not specified by user

crazy-max: > maybe we could hide it if `BUILDX_GIT_INFO=false` though

I'm not sure if this has the same purpose as `builder-id` :thinking:

.github/workflows/ci.yml
jedevc: Think we can add this as a follow-up?

jedevc (approved):

crazy-max:
@farvour

> I am using this action in our workflows to build. I am discovering that ECR does not like the attestation layers pushed when security scanning is enabled on a repository. It results in a `Failed` on the image UI state.

Is it because it cannot scan these indexes? Do you have more logs?

> Not a major issue, but sounds like this new feature to this action will let me turn these off with the newer docker buildx (0.10.0+) if my understanding is correct?

That's correct, you might need to set `provenance: false` in this case.

tonistiigi (approved):

nanake:
@crazy-max how can I properly set `provenance: false` in this case? 

```
      - name: Build
        uses: docker/build-push-action@master
        with:
          context: .
          push: true
          cache-from: type=registry,ref=ghcr.io/me/img:cache
          cache-to: type=registry,ref=ghcr.io/me/img:cache,mode=max
          tags: ghcr.io/me/img:latest
          provenance: false
```

it doesn't seem to work.

> ERROR: invalid value false
> Error: buildx failed with: ERROR: invalid value false

PR #799, opened 2023-02-09
e2e: quay.io

crazy-max:
Also adds input to set additional tag.


jedevc (approved):
LGTM, but we may want to update the repo to be under a docker controlled namespace at some point (very happy to do this as a follow-up).

crazy-max:
> but we may want to update the repo to be under a docker controlled namespace at some point (very happy to do this as a follow-up).

Sure, feel free to open a follow-up with a new slug

PR #800, opened 2023-02-10
e2e: local harbor and nexus

crazy-max:
closes #794

Test against Harbor and Nexus repositories without the need to maintain a remote instance. Was a bit of a hassle to set things up but it seems to work fine.

Also looking at https://github.com/docker/buildx/issues/1513#issuecomment-1401751716, it looks like latest stable Nexus 3.47.1 supports OCI index spec (cc @violen):

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -25,69 +25,119 @@ on:
 env:
   BUILDX_VERSION: latest
   BUILDKIT_IMAGE: moby/buildkit:buildx-stable-1
+  HARBOR_VERSION: v2.7.0
 
 jobs:
-  docker:
+  build:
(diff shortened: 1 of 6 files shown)


jedevc (approved):

.github/workflows/e2e.yml
```diff
@@ -25,69 +25,124 @@ on:
 env:
   BUILDX_VERSION: latest
   BUILDKIT_IMAGE: moby/buildkit:buildx-stable-1
+  HARBOR_VERSION: v2.7.0
```
jedevc, line 28:
```suggestion
  HARBOR_VERSION: 2.7.0
```

Would be nice if we could use the same version format for all the versions specified if possible.

PR #814, opened 2023-02-22
e2e: add distribution tests

jedevc:
:hammer_and_wrench: Fixes #812

PR #815, opened 2023-02-22
fixup! e2e: add distribution tests

jedevc:
Follow up to #814.

Ugh, sorry accidentally managed to break *all* the e2e tests :cry:


crazy-max (approved):
Ah didn't catch this for matrix input, LGTM

PR #837, opened 2023-03-12
ci: split validate and test workflow

crazy-max:
Also remove virtual-env workflow that has been moved to actions-toolkit repo and merge example workflow to ci one.


.github/workflows/ci.yml
```diff
@@ -200,6 +200,69 @@ jobs:
             exit 1
           fi
 
+  complete:
```
jedevc, line 203: Maybe `example` instead? Is the name `complete` significant?

Issue #822, opened 2023-02-27
Action not fetching images when behind HTTP proxy on self-hosted runner


grahamb:
### Troubleshooting

Before submitting a bug report please read the [Troubleshooting doc](https://github.com/docker/build-push-action/blob/master/TROUBLESHOOTING.md).

### Behaviour

When running on a self-hosted runner (RHEL8) behind a HTTP proxy, build jobs fail when attempting to fetch the Docker image referenced in the `FROM` line in the Dockerfile:

```
ERROR: failed to solve: DeadlineExceeded: DeadlineExceeded: DeadlineExceeded: node:16.19.1-alpine3.17: failed to do request: Head "https://registry-1.docker.io/v2/library/node/manifests/16.19.1-alpine3.17": dial tcp 3.216.34.172:443: i/o timeout
```

When running the same `docker buildx build` command on the runner manually with a checkout of the repo, the build succeeds.

### Configuration

* Repository URL (if public): private repo
* Build URL (if public): private repo

#### Dockerfile:
```Dockerfile
FROM node:16.19.1-alpine3.17 as build
ARG GITHUB_NPM_TOKEN
ENV NODE_ENV production
USER root
WORKDIR /usr/src/app
[logfile.txt](https://github.com/docker/build-push-action/files/10845193/logfile.txt)

COPY --chown=node:node package*.json ./
COPY --chown=node:node . .
RUN chown node:node /usr/src/app && apk update && apk add --update-cache dumb-init jq && npm -g i pm2
ARG SENTRY_DSN_ARG=""
ENV SENTRY_DSN $SENTRY_DSN_ARG
USER node
ENV NODE_JQ_SKIP_INSTALL_BINARY=true
RUN printf "//npm.github.sfu.ca/:_authToken=${GITHUB_NPM_TOKEN}\n@its:registry=https://npm.github.sfu.ca/" > .npmrc && \ 
  npm set unsafe-perm true && \
  npm config set noproxy "localhost,127.0.0.1,github.sfu.ca,*.sfu.ca,*.its.sfu.ca,docker.github.sfu.ca,www.sfu.ca" && \
  npm install --include=dev  && \
  npm run build && \
  rm /usr/src/app/.npmrc

FROM node:16.19.1-alpine3.17
WORKDIR /usr/src/app
COPY --chown=node:node --from=build /usr/src/app /usr/src/app
RUN chown node:node /usr/src/app && apk add --update-cache dumb-init jq && npm -g i pm2
USER node
[...]

PR #872, opened 2023-06-05
display proxy configuration 

crazy-max:
related to https://github.com/docker/build-push-action/issues/822
follow-up https://github.com/docker/actions-toolkit/pull/114

image


test/proxy.Dockerfile
```diff
@@ -0,0 +1,9 @@
+# syntax=docker/dockerfile:1
+FROM alpine
+RUN apk add --no-cache curl net-tools
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+RUN printenv HTTP_PROXY
+RUN printenv HTTPS_PROXY
```
thaJeztah, line 7: Just really glancing over, so no idea if relevant at all;

- both `http_proxy` (lowercase) and `HTTP_PROXY` are a thing (some software use lower, other uppercase)
- do we need some verification for `NO_PROXY` / `no_proxy` as well?

crazy-max: Both work yes: https://docs.docker.com/engine/reference/builder/#predefined-args

crazy-max: Oh you mean reading the var in the Dockerfile, let me check

crazy-max: Yes we set (upper|lower)case proxy envs: https://github.com/moby/buildkit/blob/cc2438f7ebd3fdeec4c4cb50d59ecdd6d3d9dcc5/solver/llbsolver/ops/exec.go#L383-L401

tonistiigi (approved):

PR #912, opened 2023-07-24
readme: update note/warning highlight syntax

andros21:
see https://github.com/orgs/community/discussions/16925#discussioncomment-6506860

The previous syntax should be supported for a certain period of time.
However a break space must be added between **Note** or **Warning** and first sentence.
For example:

> **Important**
>
> This doesn't work

> **Important**
> \
> This works

This PR update the two block quotes in README to use directly the new syntax:

> [!IMPORTANT]
> \
> This works (new syntax)


andros21:
Btw note/warning blocks in README are rendered correctly now
So free to stay with the previous syntax that seems to work again

PR #920, opened 2023-08-04
standardize yaml format

shubhindia:


crazy-max:
Thanks but prefer to keep as is for better readability. Same as https://github.com/docker/bake-action/pull/145#issuecomment-1611654828

shubhindia:
Ok. Makes sense but it deviates from standard structure. I mean syntactically there is nothing wrong, but most of the people would want to use the standard structure.

PR #931, opened 2023-08-16
chore: update community files

crazy-max:
Align community files.


.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Reporting security issues
+
+The project maintainers take security seriously. If you discover a security
+issue, please bring it to their attention right away!
+
+**Please _DO NOT_ file a public issue**, instead send your report privately to
+[security@docker.com](mailto:security@docker.com).
+
+Security reports are greatly appreciated, and we will publicly thank you for it.
+We also like to send gifts&mdash;if you're into schwag, make sure to let
```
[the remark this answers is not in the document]
dvdksn: Yeah I figured it was copied from somewhere. Agree that we should probably have an org-wide thing. You can keep the original text if you want, I just thought it was kind of cringe. I'll add the question about an org-wide .github repo to my backlog to investigate.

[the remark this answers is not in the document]
jedevc: Heh same issue in lots of places: https://github.com/moby/buildkit/blob/master/.github/SECURITY.md.

.github/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,4 @@
+# Code of conduct
+
+- [Moby community guidelines](https://github.com/moby/moby/blob/master/CONTRIBUTING.md#moby-community-guidelines)
+- [Docker Code of Conduct](https://github.com/docker/code-of-conduct)
```
jedevc, line 4: Hm. This repo doesn't seem to be actively updated/maintained, is this still the right thing to point to?

I'd be happy to just point to moby/moby for this.

crazy-max: Yeah indeed :disappointed: I just used the one from buildx repo. We should review that in follow-up. Removing for now.

jedevc: *super* nit picky idea :thinking: 

Could we potentially have a `make generated-files` or similar that generates these markdown files by pulling from buildx? Then we only need to maintain those in one place.

jedevc (approved):

.github/CODE_OF_CONDUCT.md
crazy-max: I think a org-wide `.github` would be better, see https://github.com/docker/build-push-action/pull/931#discussion_r1295540753

PR #947, opened 2023-09-03
Create SSH

X-oss-byte:


crazy-max:
Can you explain what's the purpose of these changes? It also looks like it's just documentation and should be done on https://github.com/docker/docs repo.

PR #948, opened 2023-09-05
feat: bump to use node20 runtime, actions/checkout to v4

kbdharun:
**Description:**

Node 16 reaches the [end of life soon on 11 Sep 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol). This PR updates the default runtime to `node20` (Node 20) (https://github.com/actions/runner/pull/2732). I have also bumped the actions/checkout version to v4 for the same.

-----------------
A major version bump might be needed after the PRs merge.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -112,6 +112,6 @@ outputs:
     description: 'Build result metadata'
 
 runs:
-  using: 'node16'
+  using: 'node20'
   main: 'dist/index.js'
   post: 'dist/index.js'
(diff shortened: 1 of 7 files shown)


action.yml
```diff
@@ -112,6 +112,6 @@ outputs:
     description: 'Build result metadata'
 
 runs:
-  using: 'node16'
+  using: 'node20'
```
crazy-max, line 115: Same as https://github.com/docker/login-action/pull/590#discussion_r1315431040

PR #954, opened 2023-09-07
chore: node 20 as default runtime

crazy-max:
follow-up
* https://github.com/docker/actions-toolkit/pull/164
* https://github.com/docker/actions-toolkit/pull/167


crazy-max:
I go ahead with this one so we can make a release today.

PR #980, opened 2023-10-04
feat: add env-secrets input

elias-lundgren:
Makes use of implementation in https://github.com/docker/actions-toolkit/commit/a1ffbe96065f5934066c01459a17967107273cc6 and the fix in https://github.com/docker/actions-toolkit/pull/176. 

A new version of the npm package (@docker/actions-toolkit) with the fix implemented in the pull request has to be released before this can be merged as I understand it.

That version should be bumped in package-lock.json/package.json as well but I don't know if dependabot or me should do it.


elias-lundgren:
@crazy-max bumped the toolkit to the rc ;)

README.md
```diff
@@ -213,38 +213,39 @@ Following inputs can be used as `step.with` keys
...
+| `cache-from`       | List        | List of [external cache sources](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from) (e.g., `type=local,src=path/to/dir`)                                                |
+| `cache-to`         | List        | List of [cache export destinations](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to) (e.g., `type=local,dest=path/to/dir`)                                              |
+| `cgroup-parent`    | String      | Optional [parent cgroup](https://docs.docker.com/engine/reference/commandline/build/#use-a-custom-parent-cgroup---cgroup-parent) for the container used in the build                                |
+| `context`          | String      | Build's context is the set of files located in the specified [`PATH` or `URL`](https://docs.docker.com/engine/reference/commandline/build/) (default [Git context](#git-context))                   |
+| `env-secrets`      | List        | List of [secrets](https://docs.docker.com/engine/reference/commandline/buildx_build/#secret) to expose to the build using environment variables from the GitHub runner (e.g., MY_SECRET=MY_ENV_VAR) |
```
crazy-max, line 228: For consistency I think it should be named `secret-envs`. It can be a `List/CSV` type as well as it should not have any special chars.

action.yml
```diff
@@ -37,6 +37,9 @@ inputs:
   context:
     description: "Build's context is the set of files located in the specified PATH or URL"
     required: false
+  env-secrets:
+    description: "List of secrets to expose to the build using environment variables from the GitHub runner (e.g., MY_SECRET=MY_ENV_VAR)"
```
crazy-max, line 41: I think "using environment variable" is enough. cc @dvdksn 

```suggestion
    description: "List of secrets to expose to the build using environment variables (e.g., MY_SECRET=MY_ENV_VAR)"
```

crazy-max (changes requested):
It seems you forgot to run `docker buildx bake pre-checkin`.

Also can you add a test in https://github.com/docker/build-push-action/blob/0f847266c302569530c95bfa228489494c43b002/__tests__/context.test.ts#L57

I was also wondering if we should check for secret id collision but can be a follow-up.

action.yml
dvdksn: agree

elias-lundgren:
> It seems you forgot to run `docker buildx bake pre-checkin`.
> 
> Also can you add a test in
> 
> https://github.com/docker/build-push-action/blob/0f847266c302569530c95bfa228489494c43b002/__tests__/context.test.ts#L57
> 
> I was also wondering if we should check for secret id collision but can be a follow-up.

I added some tests that passed here: https://github.com/docker/build-push-action/pull/980/files#diff-35112c3ed4be2816f18f19320c9c2e1fd2d10fede1dbd6ee023d75fe963ea722R622-R664 not sure if they are enough though. Also baked it as you said :)

crazy-max (changes requested):
Missing input in `action.yml`.

Also squash your commits similar to https://github.com/docker/build-push-action/pull/957/commits. Thanks.

elias-lundgren:
> Missing input in `action.yml`.
> 
> Also squash your commits similar to https://github.com/docker/build-push-action/pull/957/commits. Thanks.

Not sure what you mean with the missing input as it is added here:
https://github.com/docker/build-push-action/blob/4a0c05ce18d3e0967bdb0f790c39475ffcce40ad/action.yml#L83-L85 

I will squash the commits :)

crazy-max:
> There we go, hope it's good!

image

Sorry for the misunderstanding but I was expected two commits:
* `feat: add secrets-env input`
* `chore: update generated content` that is just the result of `docker buildx bake pre-checkin` in `./dist` like https://github.com/docker/build-push-action/pull/957/commits/ec39ef320c442d2af669fd8555e7f3b7ad7026bd as this is just generated content.

I have renamed your PR title so we are aligned.

elias-lundgren:
> > There we go, hope it's good!
> 
> image
> 
> Sorry for the misunderstanding but I was expected two commits:
> 
> * `feat: add secret-envs input`
> * `chore: update generated content` that is just the result of `docker buildx bake pre-checkin` in `./dist` like [ec39ef3](https://github.com/docker/build-push-action/commit/ec39ef320c442d2af669fd8555e7f3b7ad7026bd) as this is just generated content. See [the history](https://github.com/docker/build-push-action/commits/master/dist).
> 
> I have renamed your PR title so we are aligned.

No worries! Hopefully it's right this time.

The actions-toolkit is still on the rc.1, just wanted that known :)

crazy-max:
> The actions-toolkit is still on the rc.1, just wanted that known :)

Yes that's fine!

crazy-max (approved):
Thanks!

elias-lundgren:
Thank you!

elias-lundgren:
> how exactly does this differ from `secrets`?

@bendavies the `secrets` input takes in a list of key value pairs which are then provided as secrets to the docker build command by first writing the value to disk and referencing it, this can be seen in [`resolveBuildSecretString`](https://github.com/docker/actions-toolkit/blob/1274cd2848e2b4a9bd1aaedd19ee62a6ba3fe996/src/buildx/inputs.ts#L78-L80) which then calls [`resolveBuildSecret`](https://github.com/docker/actions-toolkit/blob/1274cd2848e2b4a9bd1aaedd19ee62a6ba3fe996/src/buildx/inputs.ts#L92-L106). The generated argument becomes `--secret id=KEY,src=<temp-file-with-secret-value>`.

`secret-envs` instead uses [`resolveBuildSecretEnv`](https://github.com/docker/actions-toolkit/blob/1274cd2848e2b4a9bd1aaedd19ee62a6ba3fe996/src/buildx/inputs.ts#L86-L90) and generates the argument `--secret id=KEY,env=ENVIRONMENT_VARIABLE` which in turn uses the `ENVIRONMENT_VARIABLE` defined on the GitHub runner as the secret value provided to the build.

Issue #958, opened 2023-09-11
Support `annotations` option


favonia:
### Description

(I am creating a new issue because the focus of the old issue #854 is unclear.)

I wish `build-push-action` can support `annotations` just like `labels`, once the new `--annotation` shortcut (https://github.com/docker/buildx/pull/2020) is officially available. Essentially, the script should append a new parameter `--annotation "line"` for each line in the value of `annotations`, just like how it appends a new `--label "line"` now for each line in the value of `labels`.

crazy-max:
`--annotation` flag PR has just been merged on buildx repo https://github.com/docker/buildx/pull/2020 and will be available on next 0.12 release. We could then add a new input for this action as follow-up. (cc @jedevc)

jedevc:
@crazy-max sgtm :tada: I think the release of v0.12 should be scheduled for fairly soon?

PR #992, opened 2023-10-24
add annotations input

crazy-max:
follow-up https://github.com/docker/buildx/pull/2020
fixes https://github.com/docker/build-push-action/issues/958


favonia:
Close #958. (Will my keyword works?)

src/context.ts
```diff
@@ -101,6 +103,11 @@ async function getBuildArgs(inputs: Inputs, context: string, toolkit: Toolkit):
       args.push('--attest', attest);
     });
   }
+  if (await toolkit.buildx.versionSatisfies('>=0.12.0')) {
+    await Util.asyncForEach(inputs.annotations, async annotation => {
+      args.push('--annotation', annotation);
+    });
+  }
```
favonia, line 110: Could we give an error in an else branch, saying it's not supported?

README.md
```diff
@@ -217,6 +217,7 @@ Following inputs can be used as `step.with` keys
 |--------------------|-------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `add-hosts`        | List/CSV    | List of [customs host-to-IP mapping](https://docs.docker.com/engine/reference/commandline/build/#add-entries-to-container-hosts-file---add-host) (e.g., `docker:10.180.0.1`)      |
 | `allow`            | List/CSV    | List of [extra privileged entitlement](https://docs.docker.com/engine/reference/commandline/buildx_build/#allow) (e.g., `network.host,security.insecure`)                         |
+| `annotations`      | List        | List of annotation to set to the image                                                                                                                                            |
```
tonistiigi, line 220: Could we get some more examples of what is the expected format?

crazy-max: @dvdksn Like other examples we have in https://docs.docker.com/build/ci/github-actions/ it might good to have one for annotations as suggested by @tonistiigi. Same in https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md for `--annotation` flag. Let me know if you need help about the syntax.

dvdksn: 👍🏻 sgtm, we only have it mentioned in the exporter docs currently I think

https://docs.docker.com/build/exporters/image-registry/#annotations
https://docs.docker.com/build/exporters/oci-docker/

tonistiigi (approved):

favonia:
@crazy-max @tonistiigi Hi, as an outsider I feel the PR might not be ready for merging. It seems to have two outstanding issues: (1) the CI was testing "v0.12.0-rc1" not the latest version, which doesn't seem to be normal, and (2) it will silently ignore annotations when buildx is too old. Of course, we can fix these in another PR, but I found the merge a bit strange. I'll be happy to learn that I'm wrong. Anyway, thank you for moving forward on this issue quickly. 😀

crazy-max:
> (1) the CI was testing "v0.12.0-rc1" not the latest version,

Yes we are going to open a follow-up here to update to latest when Buildx 0.12.0 is GA this week. We have done this quite often when adding new inputs like `build-contexts` for `0.8.0`.

PR #995, opened 2023-10-31
Update dev.Dockerfile

githraj:
build-push-action

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -45,6 +45,7 @@ if [ -n "$(git status --porcelain -- dist)" ]; then
   exit 1
 fi
 EOT
+COPY --from=format /out/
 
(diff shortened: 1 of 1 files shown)


githraj:
build-push-action

dev.Dockerfile
```diff
@@ -45,6 +45,7 @@ if [ -n "$(git status --porcelain -- dist)" ]; then
   exit 1
 fi
 EOT
+COPY --from=format /out/
```
crazy-max, line 48: This would not work and not sure what's the point of this change? We are using the `format-update` stage to bring formatted files to the client.

crazy-max:
Closing since that seems an unrelated change but feel free to bring more context if you think it should be considered.

PR #1019, opened 2023-12-01
feat: warn about ignored inputs

favonia:
Currently, several inputs can be silently ignored without any warnings. This PR will issue a warning for each ignored input with a short explanation. I actually complained about the silent input ignoring in other issues/PRs, but since this is an open-source project, I suppose I can just make a PR. 😄 If this PR is accepted, I also hope future ignore inputs will come with appropriate warnings.


src/context.ts
```diff
@@ -169,6 +175,8 @@ async function getBuildArgs(inputs: Inputs, context: string, toolkit: Toolkit):
     if (inputs.sbom) {
       args.push('--sbom', inputs.sbom);
     }
+  } else if (inputs.provenance || inputs.sbom) {
+    core.warning("Provenance is only supported by buildx >= 0.10.0; the inputs 'provenance' and 'sbom' are ignored.");
```
crazy-max, line 179:
```suggestion
    core.warning("Attestations are only supported by buildx >= 0.10.0; the inputs 'provenance' and 'sbom' are ignored.");
```

crazy-max (changes requested):
Minor nit otherwise LGTM

src/context.ts
favonia: Thanks! Fixed in my local copy. (Will squash everything into a commit later.)

favonia:
@crazy-max I believe all issues were addressed. 😀

PR #1049, opened 2024-02-03
Env var to set provenance mode

crazy-max:
When building from a private repository, provenance will be set to `mode=min` by default: https://github.com/docker/build-push-action/blob/94d76d3bc1409736cb5dc1ada9502bec3a72973c/src/context.ts#L166-L169

And with the right `builder-id` such as:

```
mode=min,inline-only=true,builder-id=https://github.com/foo/bar/actions/runs/7767303517
```

In some cases we might want to set it to mode max but keep the generated builder id. Currently we need to set `provenance` input to something like this in a workflow to make it consistent with the toolkit behavior:

```yaml
      -
        name: Build and push
        uses: docker/build-push-action@v5
        with:
          push: true
          provenance: mode=max,builder-id=${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
          tags: user/app:latest
```

This is similar to what is returned when building from a public repository: https://github.com/docker/build-push-action/blob/94d76d3bc1409736cb5dc1ada9502bec3a72973c/src/context.ts#L170-L172

This change handles an env var called `BUILDX_PROVENANCE_MODE` that can be one of `auto`, `min`, `max` and defaults to `auto` to keep the current behavior. This avoids to set the builder-id:

```yaml
      -
        name: Build and push
        uses: docker/build-push-action@v5
        with:
          push: true
          tags: user/app:latest
        env:
          BUILDX_PROVENANCE_MODE: max
```

Alternatively we could support handlebar templates for provenance input with a global expression `{{workflow_run_url}}`:

```yaml
      -
        name: Build and push
        uses: docker/build-push-action@v5
        with:
          push: true
          provenance: mode=max,builder-id={{workflow_run_url}}
          tags: user/app:latest
```


tonistiigi (commented):
Instead of adding a new field (that control something we already have a field for) can't we just detect if user has set `builder-id` in the provenance settings and if they have not then add the default?

crazy-max:
> Instead of adding a new field (that control something we already have a field for) can't we just detect if user has set `builder-id` in the provenance settings and if they have not then add the default?

We already do this actually in our toolkit :man_facepalming:: https://github.com/docker/actions-toolkit/blob/3f7d7f75b525f1e923d54c0ce95a255f89218965/src/buildx/inputs.ts#L122-L143

I was just confused when using the bake action because I was using `set: *.attest=type=provenance,mode=max` which is not covered by the toolkit. Using `provenance: mode=max` in `bake-action` works just fine. Closing this one.

PR #1086, opened 2024-03-26
handle attests correctly with provenance and sbom inputs

crazy-max:
We don't detect usage of `attests` input and therefore setting `attests: type=provenance,mode=max` will result in a duplicated provenance as we already set it automatically and build will fail with: 

```
ERROR: duplicate attestation field provenance
```

This will also set `builder-id` automatically if `provenance` is used in `attests` which was not the case before.

To keep everything consistent across our flags such as `load` and `push`, `provenance` and `sbom` inputs take precedence over ones set in `attests`


src/context.ts
```diff
@@ -238,3 +215,69 @@ async function getCommonArgs(inputs: Inputs, toolkit: Toolkit): Promise<Array<st
...
+  });
+  for (const attr of attributes) {
+    for (const [key, value] of attr.map((chunk: string) => chunk.split('=').map(item => item.trim()))) {
+      if (key == 'type' && value == name) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
```
crazy-max, line 283: keeping in draft for now as this func should be implemented in the toolkit.

crazy-max:
https://github.com/docker/build-push-action/actions/runs/8437957902/job/23108984199?pr=1086#step:5:123

```
/usr/bin/docker buildx build --cache-from type=gha,scope=attests-image --cache-to type=gha,scope=attests-image,mode=max --file ./test/go/Dockerfile --iidfile /home/runner/work/_temp/docker-actions-toolkit-3KzNXb/iidfile --output type=image,name=localhost:5000/name/app:latest,push=true --attest type=provenance,mode=max,builder-id=https://github.com/docker/build-push-action/actions/runs/8437957902 --attest type=sbom,true --target image --metadata-file /home/runner/work/_temp/docker-actions-toolkit-3KzNXb/metadata-file ./test/go
ERROR: invalid value true
```

We need the canonical representation when using `--attest` similar to https://github.com/docker/buildx/blob/8abef5908705e49f7ba88ef8c957e1127b597a2a/util/buildflags/attests.go#L13-L21

tonistiigi (approved):

PR #1114, opened 2024-05-10
ci(e2e): update artifactory instance

crazy-max:
cc @thompson-shaun @shrug


crazy-max:
Getting this one in to check workflow

PR #1267, opened 2024-11-26
iterate over allow input values and push each separately

crazy-max:
To be consistent with other inputs.


crazy-max:
> That might be a bug when parsing this flag in buildx but we can mitigate this in the action by passing the flag n time.

Ah my bad `fs` entitlement is not a thing for `build` command. It's only for `bake` one. But this change is still useful so we have a better readable command in the logs.

PR #1321, opened 2025-02-16
reusable workflow to distribute multi-platform builds

crazy-max:
similar to https://github.com/docker/bake-action/pull/302


crazy-max:
Switch to https://github.com/docker/github-builder

Issue #1334, opened 2025-03-03
Uploaded artifact are not zipped


toots:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug, and:

- [x] The documentation does not mention anything about my problem
- [x] There are no open or closed issues that are related to my problem

### Description

Hi!

I looked for previous issues and did not find any. Apologies if I missed one.

Our build system had become flakey recently after being rock solid for a while. After tinkering and looking around, I think that I have traced it to this action.

The problem seems to be that the artifact uploaded for the docker build summary are improperly described as zip files but are in fact not compressed.

Here are a couple of breadcrumbs:
* Issue comment explaining that `actions/download-artifact` fails on non-zip files: https://github.com/actions/download-artifact/issues/328#issuecomment-2199544827
* Relevant code from this action:
```typescript
          core.info(`Build record written to ${exportRes.dockerbuildFilename} (${Util.formatFileSize(exportRes.dockerbuildSize)})`);

          let uploadRes: UploadArtifactResponse | undefined;
          if (recordUploadEnabled) {
            uploadRes = await GitHub.uploadArtifact({
              filename: exportRes.dockerbuildFilename,
              mimeType: 'application/gzip',
              retentionDays: recordRetentionDays
            });
          }
```
no trace of any zipping process here.
* Faulty artifact download logs: https://github.com/savonet/liquidsoap/actions/runs/13503019183/job/38070586645

You will find it pretty hard to know which artifact is failing in the last log but here's the final smoking gun:
* Uploaded docker build summary: https://github.com/savonet/liquidsoap/actions/runs/13591496099/attempts/1

Try downloading any of the docker build summary artifact in this page. You get a file named `savonet~liquidsoap~1UVFX2.dockerbuild.zip` for which:
```shell
% unzip savonet\~liquidsoap\~1UVFX2.dockerbuild.zip                                                                                                                                                     18:56:53
Archive:  savonet~liquidsoap~1UVFX2.dockerbuild.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of savonet~liquidsoap~1UVFX2.dockerbuild.zip or
[...]

Issue #1167, opened 2024-07-01
v6 cannot be used with actions/download-artifact@v4


kateinoigakukun:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug, and:

- [X] The documentation does not mention anything about my problem
- [X] There are no open or closed issues that are related to my problem

### Description

After updating `docker/build-push-action` to v6, we started observing the following error at `actions/download-artifact` step.

```
Unable to download artifact(s): Unable to download and extract artifact: Artifact download failed after 5 retries.
```

I guess something in the build summary uploading feature is incompatible with `actions/download-artifact`.

You can reproduce the problem with this  https://github.com/kateinoigakukun/gh-actions-sandbox/actions/runs/9737812232/workflow

### Expected behaviour

`actions/download-artifact` can download all artifacts even with this action.

### Actual behaviour

`actions/download-artifact` fails while trying to download all artifacts

### YAML workflow

```yaml
name: Docker Build

on:
  workflow_dispatch:

jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      - uses: docker/build-push-action@v6
  post-job:
    runs-on: ubuntu-latest
    needs: [docker]
    steps:
      - uses: actions/download-artifact@v4
```

crazy-max:
Thanks for your feedback, looking at the workflow:

```yaml
name: Docker Build

on:
  workflow_dispatch:

jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      - uses: docker/build-push-action@v6
  post-job:
    runs-on: ubuntu-latest
    needs: [docker]
    steps:
      - uses: actions/download-artifact@v4
```

It seems `actions/download-artifact@v4` downloads all available artifacts within the workflow if `name` or `pattern` is not set: https://github.com/actions/download-artifact?tab=readme-ov-file#download-all-artifacts

> If the `name` input parameter is not provided, all artifacts will be downloaded. To differentiate between downloaded artifacts, by default a directory denoted by the artifacts name will be created for each individual artifact. This behavior can be changed with the `merge-multiple` input parameter.

And will therefore try to download the `.dockerbuild` file uploaded by our action.

```
Starting download of artifact to: /home/runner/work/gh-actions-sandbox/gh-actions-sandbox/kateinoigakukun~gh-actions-sandbox~8P1DEG.dockerbuild
[...]

PR #1337, opened 2025-03-07
note about usage of summary feature with download-artifact action

crazy-max:
relates to
* https://github.com/actions/download-artifact/issues/367
* https://github.com/actions/download-artifact/issues/328#issuecomment-2199704064
* https://github.com/docker/build-push-action/issues/1167
* https://github.com/docker/actions-toolkit/issues/490
* https://github.com/docker/bake-action/issues/280
* https://github.com/docker/build-push-action/issues/1334
* https://github.com/actions/toolkit/pull/1874

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,19 @@ additional details about the build, including build stats, logs, outputs, and
 more. The build record can be imported to Docker Desktop for inspecting the
 build in greater detail.
 
+> [!WARNING]
+>
+> If you're using the [`actions/download-artifact`](https://github.com/actions/download-artifact)
+> action in your workflow, you need to ignore the build record artifacts
+> if `name` and `pattern` inputs are not specified ([defaults to download all artifacts](https://github.com/actions/download-artifact?tab=readme-ov-file#download-all-artifacts) of the workflow),
+> otherwise the action will fail:
+> ```yaml
+> - uses: actions/download-artifact@v4
+>   with:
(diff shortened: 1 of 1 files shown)


thompson-shaun (approved):

README.md
```diff
@@ -185,6 +185,19 @@ additional details about the build, including build stats, logs, outputs, and
 more. The build record can be imported to Docker Desktop for inspecting the
 build in greater detail.
 
+> [!WARNING]
+>
+> If you're using the [`actions/download-artifact`](https://github.com/actions/download-artifact)
+> action in your workflow, you need to ignore the build record artifacts
+> if you don't specify the `name` input (download all artifacts), otherwise the
```
tonistiigi, line 192: What is "name" input? You mean "pattern" ?

crazy-max: No I really mean `name`: https://github.com/actions/download-artifact/blob/cc203385981b70ca67e1cc392babf9cc229d5806/action.yml#L5-L7

Maybe I can just remove this part if that's confusing

tonistiigi: Ah, it is confusing because the example is using `pattern`. If we want to be very clear, I think we should provide examples for both name and pattern.

crazy-max: Updated to explain that this applies only if name and pattern are not set.

tonistiigi (approved):

PR #1341, opened 2025-03-27
Clarify .dockerignore usage in README

npwolf:
The previous instructions made me think mutation of the .dockerignore file was not supported with the git context, but the the basic .dockerignore was supported (not the case). Just trying to improve the doc so it is more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ jobs:
 ```
 
 Be careful because **any file mutation in the steps that precede the build step
-will be ignored, including processing of the `.dockerignore` file** since
+will be ignored. Additionally the `.dockerignore` file is not used and can only be utilized by using path context.** since
 the context is based on the Git reference. However, you can use the
 [Path context](#path-context) using the [`context` input](#inputs) alongside
 the [`actions/checkout`](https://github.com/actions/checkout/) action to remove


README.md
```diff
@@ -81,7 +81,7 @@ jobs:
 ```
 
 Be careful because **any file mutation in the steps that precede the build step
-will be ignored, including processing of the `.dockerignore` file** since
```
crazy-max, line 84: I think this is already self-explanatory that `.dockerignore` is ignored with Git context here. We also say that Path context should be used in next sentence if you don't want this behavior.

crazy-max (commented):
> but the the basic .dockerignore was supported (not the case)

Not sure I understand this statement. Do you mean `.dockerignore` file within a Git context is supported?

README.md
npwolf: @crazy-max I personally thought the text was saying file mutations in general and file mutations of the .dockerignore file specifically weren't supported with git context (not that .dockerignore wasn't supported at all in git context). I've updated this with another revision to try to improve the clarity for others. Let me know if I can improve it further. If you don't feel it warrants an update, feel free to close the PR. Thanks!

mschoettle:
I have a `Dockerfile` that expects `.git` to be present. So I ran into this issue with the git vs path context. Once I found the relevant part in the README the explanation there was clear to me.

The problem I had is that I looked in the wrong place of the README (the examples section). So maybe we could add a link to this part of the README in the examples section?

crazy-max:
> I have a `Dockerfile` that expects `.git` to be present. So I ran into this issue with the git vs path context. Once I found the relevant part in the README the explanation there was clear to me.
> 
> The problem I had is that I looked in the wrong place of the README (the examples section). So maybe we could add a link to this part of the README in the examples section?

You can keep the `.git` folder with Git context if you set the build-arg `BUILDKIT_CONTEXT_KEEP_GIT_DIR=1`: https://docs.docker.com/reference/dockerfile/#buildkit-built-in-build-args

PR #1343, opened 2025-03-28
handle no default attestations env var

### LINKED ISSUE #1339: `BUILDX_NO_DEFAULT_ATTESTATIONS` not respected after v5.3.0
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug, and:

- [x] The documentation does not mention anything about my problem
- [x] There are no open or closed issues that are related to my problem

### Description

We have been successfully building multi arch images with native GHA runners. But when updating `build-push-action` we noticed that `BUILDX_NO_DEFAULT_ATTESTATIONS=1` no longer works in latest version of the action (v6.x). I was able to find that this got broken already in v5.3.0. We were previously using v5.0.0 so the jump to the latest version was quite large version bump.

(We on purpose do not use any QEMU based approaches for building the images. These are also private images we are building.)

### Expected behaviour

`docker manifest create` should work in the example YAML workflow.

### Actual behaviour

`docker manifest create` breaks with the cryptic error `myregistry/myrepo:mytest-amd64 is a manifest list`. This means that the arch specific images got some attestation data in them although `BUILDX_NO_DEFAULT_ATTESTATIONS=1` was used as [documented](https://docs.docker.com/build/building/variables/#buildx_no_default_attestations).

### Repository URL

_No response_

### Workflow run URL

_No response_

### YAML workflow

```yaml
name: Build multi arch image

on: workflow_call

jobs:
  test-multi-arch-build:
    strategy:
      matrix:
        include:
          - runner: ubuntu-latest
            suffix: amd64
          - runner: linux_ubuntu_24_04_arm64_2_cores
            suffix: arm64
    runs-on: ${{ matrix.runner }}
    env:
      BUILDX_NO_DEFAULT_ATTESTATIONS: 1  # Buildx added attestations are not compatible with "docker manifest" cmds
    steps:
      - uses: actions/checkout@v4.2.2

      - uses: docker/setup-buildx-action@v3.10.0

      - name: Build push image
        uses: docker/build-push-action@v5.3.0  # 5.3.0 is broken but 5.2.0 works. Also 6.x are broken.
        with:
          context: .
          tags: "myregistry/myrepo:mytest-${{ matrix.suffix }}"
          push: true

  test-multi-arch-manifest:
    needs: [ test-multi-arch-build ]
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2

      - name: Create Docker multi arch manifest
        env:
          IMAGE: "myregistry/myrepo:mytest"
        run: docker manifest create $IMAGE --amend $IMAGE-amd64 --amend $IMAGE-arm64
```

### Workflow logs
Error: `myregistry/myrepo:mytest-amd64 is a manifest list` from docker manifest create

crazy-max:
fixes #1339 

We should not set provenance attestation if [`BUILDX_NO_DEFAULT_ATTESTATIONS` env var](https://docs.docker.com/build/building/variables/#buildx_no_default_attestations) is set.


tonistiigi (commented):
Is the issue that env is not propagated to buildx or does this need custom handling? If former then maybe a better fix would be to just make sure all `BUILDX_*` env are forwarded.

crazy-max:
> or does this need custom handling

Yes it needs custom handling because the action is setting the provenance based on github events: https://github.com/docker/build-push-action/blob/c566248492c912e39910ac79e2f05a82260233a8/src/context.ts#L248-L260

tonistiigi (approved):

PR #1344, opened 2025-04-01
refactor: add missing 'new' and exception handling

k-kbk:
- Synchronization methods such as `fs.rmSync(...)` that throw an exception on failure are [recommended](https://nodejs.org/api/fs.html#synchronous-example) to be wrapped in try/catch.
- Using `new` to create an object is the standard and safe way to do it.


crazy-max (dismissed):
~LGTM thanks!~

crazy-max (changes requested):
Can you remove that last empty commit please? Otherwise LGTM

crazy-max (approved):
Thanks

PR #1351, opened 2025-04-07
enforce secrets input value as registered secret

crazy-max:
follow-up https://github.com/docker/actions-toolkit/pull/645

Before:

image

After:

image


crazy-max:
As discussed internally we can just show the keys in build summary.

PR #1361, opened 2025-04-17
docs: add validating build configuration example

mschoettle:
Follow-up from #1359 to add a reference to the validating build configuration documentation.


crazy-max (approved):
LGTM thanks!

PR #1370, opened 2025-04-30
feat: add support for Docker Build Cloud summary

nico1510:
## What’s Changed

- Adds support for Docker Build Cloud: when the `cloud` driver is used, the action now generates a summary with a direct link to the build details on app.docker.com.

**Test:**  
Run the action with both drivers and check the summary output for the correct link and format.

Example output:
image

image


src/main.ts
```diff
@@ -222,6 +223,18 @@ actionsToolkit.run(
...
+      if (platform && refId) {
+        const buildUrl = `https://app.docker.com/build/accounts/docker/builds/${platform}/${refId}`;
+
+        core.info(`View build details: ${buildUrl}`);
+
+        const sum = core.summary.addHeading('Docker Build Cloud summary', 2);
+        sum.addRaw('<p>').addRaw('Your build was executed using Docker Build Cloud. ').addRaw('You can view detailed build information, logs, and results here: ').addLink(buildUrl, buildUrl).addRaw('</p>');
+        sum.addRaw('<p>').addRaw('For more information about Docker Build Cloud, see ').addLink('the documentation', 'https://docs.docker.com/build/cloud/').addRaw('.').addRaw('</p>');
+        await sum.addSeparator().write();
+      }
```
crazy-max, line 237: Related to previous comment we can remove this condition and we should move this logic to the actions-toolkit in `GitHub.writeBuildSummary`: https://github.com/docker/actions-toolkit/blob/ba0e8d7ed8b86856c1605c90b57f7ee543d6740b/src/github.ts#L229 so it can be used with `bake-action` as well.

Btw how multiple builds with bake would be handled?: https://github.com/docker/bake-action/blob/76f9fa3a758507623da19f6092dc4089a7e61592/src/main.ts#L210-L246

We need to be consistent between both `build-push-action` and `bake-action` imo. I think for multiple builds with bake we could have the list of builds filtered like `https://app.docker.com/build/accounts/docker/builds?refs=foo,bar` if possible?

crazy-max (commented):
With https://github.com/docker/build-push-action/pull/1364 we could have both export and link to the build for cloud driver I think.

crazy-max:
fixed in https://github.com/docker/build-push-action/pull/1381

PR #1374, opened 2025-05-12
Explain can't push tagged ref error

jedahan:
The error got me hunting for an underlying explanation.


TROUBLESHOOTING.md
```diff
@@ -135,3 +135,11 @@ Or a dedicated step to sanitize the slug:
     push: true
     tags: ${{ steps.repo_slug.outputs.result }}:latest
 ```
+
+## can't push tagged ref docker.io/user/image:latest by digest
```
crazy-max, line 139: Is this the actual error, can you provide repro and logs please?

PR #1418, opened 2025-09-18
Create devcontainer.json

mikeb42069:
zacharyr0th:patch-5


crazy-max:
I don't think there's any point in using a dev container for this project. Can you expand on your use case?

PR #1422, opened 2025-10-28
feat: Adds retry functionality to handle transient build failures with configurable parameters.

morwn:
### Description
Add retry mechanism with configurable maximum attempts, wait time, and timeout

### New input parameters:
- max-attempts (default: 1) - Maximum number of build attempts
- retry-wait-seconds (default: 0) - Delay between retry attempts
- timeout-minutes (default: 0) - Timeout per attempt (0 = no timeout)

### Implementation:
- Wraps build execution in retry loop with comprehensive logging
- Adds timeout support per attempt using Promise.race()
- Fully backward compatible (default values maintain current behavior)
- Adds 2 test cases


crazy-max:
Thanks for taking the time to implement this. We're declining it because it changes responsibility boundaries in this action; retry/failure policy should be handled upstream rather than in the GitHub Action wrapper. Closing.

morwn:
Thanks for the explanation. However, "handle it upstream" has no clean implementation path. GitHub Actions provides no native mechanism to retry a uses: step - only run steps can be wrapped by retry actions like nick-fields/retry. The only alternatives are duplicating the entire job or bypassing build-push-action entirely with raw docker buildx build + docker push shell commands.

Transient failures (registry timeouts, BuildKit connectivity issues) are a real and frequent pain point for users of this action specifically. Keeping retry logic here allows it to be scoped correctly to the build attempt itself, without duplicating setup overhead.

The retry logic can't be moved upstream without either significant workflow complexity or losing the abstraction this action provides.

crazy-max:
Thanks for the follow-up.                                                                                                                                                                                                                                                                                                                        

You're correct that GitHub Actions does not currently provide a clean native retry for `uses:` steps (something they wanted to implement some time ago iirc), and I understand the pain from transient connectivity failures.

We're still going to keep this PR closed because adding retry policy inside build-push-action would move this action from "build invocation" into workflow orchestration/failure policy, which we intentionally keep outside this repository. That boundary is important for predictability and long-term maintenance.
                                                                                                                                                                                                                                                                                                                                               
We agree the upstream options are imperfect today, but we prefer that tradeoff over embedding retry semantics in this wrapper.

Issue #1092, opened 2024-04-06
Add support for --progress CLI option


TheFriendlyCoder:
### Description

The Docker CLI provides an option --progress for customizing the verbosity level produced by the tool. This is particularly helpful when building from a CI because some of the default status messages produced by the tool. For example, when pulling or pushing containers Docker will, by default, show a percentage status update / progress bar to the user. This is great when running within an interactive shell but it just produces unnecessary noise in a build log when running on CI.

It'd be nice if you exposed this property to your GH action plugin so we could easily control the verbosity in a similar way. Perhaps this could be as simple as adding a customization parameter to your plugin config with the same name. It could be optional to preserve backwards compatibility.

crazy-max:
There is no tty on GitHub runners afaik so doesn't make sense imo. That's why the current progress in GitHub actions defaults to `plain` otherwise would be `tty` if detected.

But you can still use the [`BUILDKIT_PROGRESS`](https://docs.docker.com/build/building/variables/#buildkit_progress) env var if you want:

```yaml
      -
        name: Build and push
        uses: docker/build-push-action@v5
        with:
          push: true
          tags: user/app:latest
        env:
          BUILDKIT_PROGRESS: tty
```

Issue #1014, opened 2023-11-23
Plain output for buildx ?


azr:
### Description

Howdy !
Sometimes, when building with buildx, things fail, but the output of the things happening in the docker build will be clipped, which can make it hard to know what happened from the build logs.

I'm proposing to add way to a `--progress plain` option to a build.

Either by passing a map of argvs, or by adding a new option.

Thanks !

Edit: I can do it if you'd like.

azr:
Ah, wait, the `BUILDKIT_PROGRESS` variable could be used for this, nvm.

Issue #885, opened 2023-06-20
Enable Buildx progress=plain


josherickson:
Hi there, wondering if it's possible to enable the `progress=plain` build setting. It's described [here](https://docs.docker.com/engine/reference/commandline/buildx_build/#progress) and looks like it can either be set with an argument `--progress=plain` in the build command, or as an environment variable `BUILDKIT_PROGRESS=plain`.

Have tried setting this an environment variable on the action but it didn't seem to have any effect.

```
- name: Build and push Docker image
  uses: docker/build-push-action@v4
  env:
    BUILDKIT_PROGRESS: plain
  with:
    build-args: ${{ inputs.docker-build-args }}
    context: ${{ inputs.docker-context }}
    file: ${{ inputs.docker-context }}/${{ inputs.dockerfile }}
    push: true
    tags: latest
```

Thanks!

crazy-max:
There is no `tty` available on GitHub runners and therefore defaults to `plain` such as https://github.com/docker/build-push-action/actions/runs/5321201118/jobs/9636009612#step:6:114

josherickson:
Ok thanks @crazy-max, I think our expectations of what that does might've been wrong then. We were hoping to get a simpler output than what our GitHub Actions build logs look like. For example, below we have the same `#17 [10/11] RUN ...` command three times with all the extracting logs mixed in. Thought that this might be a setting that would make this easier to follow.

```
#17 [10/11] RUN chmod 700 node_modules/puppeteer/.local-chromium/linux-*/chrome-linux/chrome
#17 sha256:35eb1fdebbb4820d0127f245e009a1f43bf2d27e28dce06f10a72fe72dbc7698 0B / 101.05MB 0.2s
#17 sha256:54682432b4b7f3c6abb20ea929649511f0de6081d4e3ec4e295c43f1fe[396](https://github.com/BetterOfficeApps/core-api/actions/runs/5328102425/jobs/9652337083#step:4:412)40b 0B / 141.25MB 0.2s
#17 sha256:9eaf03f74596eac4c833d95b7d4fc93c90bc2dc3ecfd31612e25a66b3b6f1d25 5.81kB / 5.81kB 0.3s done
#17 sha256:07c52888fa9c2e059c47a5111fa2451407d85a78992891d10b48cea21b32553b 0B / 844.45MB 0.2s
[...]

PR #1423, opened 2025-10-29
feat: add quiet input that maps to --quiet

szeyu:
This pull request introduces support for a new `quiet` input, allowing users to suppress build output by mapping to the `--quiet` flag. The changes update the documentation, action input definitions, input parsing, and argument construction to support this new feature.

**Support for `quiet` build option:**

* Added `quiet` input to `action.yml` with description and default value, enabling users to suppress build output.
* Documented the new `quiet` input in the `README.md` table of inputs.

**Code updates for input handling:**

* Added `quiet` property to the `Inputs` interface in `src/context.ts`.
* Updated `getInputs()` in `src/context.ts` to parse the `quiet` input as a boolean.
* Modified `getCommonArgs()` in `src/context.ts` to append `--quiet` to build arguments if the input is set.


crazy-max:
Thanks for your contrib but similar to:
* https://github.com/docker/build-push-action/issues/1092
* https://github.com/docker/build-push-action/issues/1014
* https://github.com/docker/build-push-action/issues/885#issuecomment-1599693392

We don't want to provide explicit input as it can be done through [`BUILDKIT_PROGRESS` env var](https://docs.docker.com/build/building/variables/#buildkit_progress). In your case you can do:

```yaml
      -
        name: Build and push
        uses: docker/build-push-action@v6
        with:
          push: true
          tags: user/app:latest
        env:
          BUILDKIT_PROGRESS: quiet
```

You also have `none` available since Buildx 0.29 that doesn't print the image ID.

PR #1448, opened 2026-01-24
Add parameters for controlling inheritance of labels and annotations

seschi98:
Add new parameters `inherit-annotations` and `inherit-labels` that allows disabling inheritance from the base image or base states.


seschi98:
Well, wenn testing this changes in my workflow I realised those parameters dont't actually exist in the buildx CLI. I saw them on my local machine because I have setup podman as a docker alias, so epic fail from my side 🤦‍♂️

PR #1554, opened 2026-06-02
ci(e2e): use GITHUB_TOKEN for GHCR e2e

crazy-max:
Updates the e2e workflow so GHCR authentication uses the workflow `GITHUB_TOKEN` instead of a stored PAT.

The caller workflow grants `packages: write` to the reusable e2e job, and the reusable workflow falls back to `github.actor` and `secrets.GITHUB_TOKEN` when the registry is `ghcr.io`.

This removes the need for a long-lived GHCR PAT while keeping the generic registry secret mapping focused on external registries.

### DIFF
--- a/.github/workflows/.e2e-run.yml
+++ b/.github/workflows/.e2e-run.yml
@@ -1,9 +1,6 @@
 # reusable workflow
 name: .e2e-run
 
-permissions:
-  contents: read
-
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/.e2e-run.yml
```diff
@@ -1,9 +1,6 @@
 # reusable workflow
 name: .e2e-run
 
-permissions:
```
thaJeztah, line 4: why was this removed?

.github/zizmor.yml
```diff
@@ -0,0 +1,6 @@
+rules:
+  # rule does not apply to reusable worfklows where permissions are defined by
```
thaJeztah, line 2: ah, right; so wouldn't it still make sense to set defaults, or won't that work?

.github/workflows/.e2e-run.yml
crazy-max: Because we introduce `packages: write` and caller should take care of permissions, not the reusable workflow. There is some analysis here: https://github.com/docker/github-builder/issues/24

.github/zizmor.yml
crazy-max: Yeah that won't work unfortunately, see https://github.com/docker/github-builder/issues/24

thaJeztah: That's annoying! Is it a known issue in zizmor? Is this something that could be fixed in GitHub actions itself?

thaJeztah (approved):
LGTM

.github/zizmor.yml
crazy-max: I think this is tracked inb zizmor yeah

PR #1561, opened 2026-06-10
ci(e2e): use OIDC for AWS ECR

crazy-max:
The e2e workflow now configures AWS credentials through GitHub OIDC for AWS ECR and ECR Public jobs.

This removes the AWS access key secrets from the registry credential matrix while keeping the other registry credentials unchanged.

cc @zampani-docker

### DIFF
--- a/.github/workflows/.e2e-run.yml
+++ b/.github/workflows/.e2e-run.yml
@@ -10,6 +10,9 @@ on:
       type:
         required: true
         type: string
+      provider:
+        required: true
+        type: string
       name:
         required: true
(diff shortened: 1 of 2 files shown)


crazy-max:
Expected to fail for now:

```
Assuming role with OIDC
Retry AssumeRole: attempt 1 of 12 failed: Could not assume role with OIDC: Not authorized to perform sts:AssumeRoleWithWebIdentity. Retrying after 17ms.
```

tonistiigi (approved):

